PR #1, opened 2021-04-01
PAAS-1508 first dirty push which stains

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1508

Short description:


packages/one-shot/check_jexperience.yml
```diff
@@ -0,0 +1,41 @@
...
+type: update
+version: 1.5.2
+name: Jahia - check jexperience
+id: jahia-check-jexperience
+description:
+  short: Jahia - check jexperience
+
+mixins:
+  - ../../mixins/jahia.yml
+  - ../../mixins/jcustomer.yml
```
kevinvandel, line 11: We should "import" `common.yml` too, even though we don't use actions from `jahia.yml` or `jcustomer.yml` that really requires it. But since it's marked as a dependency, I think we should keep it stupid and nasty and systematically importing `common.yml` when importing another mixin that depends on it.

README.md
```diff
@@ -1,3 +1,211 @@
...
+│   │   └── upgrade.yml
+│   ├── jcustomer
+│   └── one-shot
+│       ├── check_jexperience.yml
+│       ├── ipsec.yml
+│       ├── manage-auth-basic.yml
+│       ├── reset-polling.yml
+│       ├── rewrite-rules.yml
+│       ├── set-jahia-root-password.yml
+│       └── set-jahia-tools-password.yml
```
kevinvandel, line 69: To me all those packages should be moved to `jahia` folder, except for `reset-polling.yml` which should be moved to `common` folder.

About the "Structure" section of the readme, I think we should only put the folders structure, put not the files (it would be boring to update each time we add a new package).

kevinvandel (changes requested):

packages/one-shot/check_jexperience.yml
laurentfufu: done !

README.md
laurentfufu: move done (readme updated)
the tree have been updated with directories only

kevinvandel (approved):
I will let someone else have a look before merging
@Ronan-Gorain if you have time for that

assets/common/set_dd_tags.sh
```diff
@@ -0,0 +1,44 @@
...
+echo "  processing_rules:"
+echo "    - type: mask_sequences"
+echo "      name: mask_password"
+echo "      pattern: (PASSWORD=)(\w+)"
+echo "      replace_placeholder: \"PASSWORD=[masked_password]\""
+echo "tags:"
+echo " - product:$(awk '$2~/^product/ {split($2,a,":"); print a[2]}' $ddconffile)"
+echo " - envname:${envName}"
+echo " - provide:${_PROVIDE}"
+echo " - version:${DX_VERSION}"
```
Ronan-Gorain, line 24: This variable is specific to jahia. 
Do we plan to update the script when merging other repos ? Otherwise, it should be place in assets/jahia until we refactor it

Ronan-Gorain: Just talk with Kevin, this script is only used by haproxy and mariadb nodes, for all others, it's in the jahiastic image. So we don't care.

PR #5, opened 2021-04-09
PAAS-1512 Move cloud-scripts Jelastic packages to jelastic-packages repo

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1512

Short description:


packages/augsearch/augsearch-redeploy-es-nodes.yml
```diff
@@ -0,0 +1,30 @@
...
+mixins:
+  - ../../mixins/common.yml
+  - ../../mixins/jahia.yml
+  - ../../mixins/elasticsearch.yml
+
+onInstall:
+  - setGlobalRepoRootUrl
+  - enableFullReadOnlyOnCluster
+  - checkEsClusterStatus
+  - redeployEsNodes
```
kevinvandel, line 21: This will not work, because those 2 actions needs to be performed on the AS env, but here they'll be run on the Jahia env.
That's why in the original package I did it like that:
```
  - install:
      jps: "${baseUrl}/../es-nodes/redeploy-es-nodes.yml"
      envName: ${settings.augSearchEnvName}
```

README.md
```diff
@@ -99,8 +99,29 @@ Once it is completed, you need to apply `jahia/perf-test-step2.yml` package to t
 
 ## Packages
 
+### augsearch
+
+#### augsearch/augsearch-redeploy-es-nodes.yml
```
kevinvandel, line 104: If we really wanted to comply with the rules, we should put this package inside the "jahia" folder, since it needs to be applied to the Jahia env (and not the AS one, which is given in a separate custom argument).
But since this one is really particular, maybe we can leave it here for now (we'll have to redo all this when integrating AS for real anyway).

kevinvandel (changes requested):

packages/augsearch/augsearch-redeploy-es-nodes.yml
laurentfufu: as requested, the manifest is now deleted

README.md
laurentfufu: as requested, the manifest is now deleted

kevinvandel (approved):

Ronan-Gorain (approved):

PR #6, opened 2021-04-09
PAAS-1509 fix empty password setting not working and jcustomer upgrade

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1509

Short description:

### DIFF
--- a/packages/jcustomer/update-events.yml
+++ b/packages/jcustomer/update-events.yml
@@ -55,8 +55,11 @@ onBeforeRedeployContainer[es]:
 onAfterRedeployContainer[cp]:
   - getUnomiRootPassword
   - setReplica
-  - setupUnomi
-  - setupDatadogAgentUnomi: cp
+  - setupUnomi:
+      unomi_root_password: ${globals.unomi_root_password}
+  - setupDatadogAgentUnomi:
+      target: cp
+      unomi_root_password: ${globals.unomi_root_password}
   - if ('${event.params.nodeGroup.print()}' != ''):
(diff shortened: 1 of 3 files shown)


packages/jcustomer/update-events.yml
```diff
@@ -55,8 +55,11 @@ onBeforeRedeployContainer[es]:
 onAfterRedeployContainer[cp]:
   - getUnomiRootPassword
   - setReplica
-  - setupUnomi
-  - setupDatadogAgentUnomi: cp
+  - setupUnomi:
+      unomi_root_password: ${globals.unomi_root_password}
+  - setupDatadogAgentUnomi:
+      target: cp
+      unomi_root_password: ${globals.unomi_root_password}
```
kevinvandel, line 62: Won't it be an issue when applying the `update-events` package directly ? (so not from the `install` package, and so the globals won't be defined)
I checked how it's done in `paas_jelastic_unomi`, but I don't understand why it works...

laurentfufu: This should be ok because onBefore event use action `saveUnomiRootPassword`: https://github.com/Jahia/jelastic-packages/blob/PAAS-1509/packages/jcustomer/update-events.yml#L37-L38
and onAfter event use `getUnomiRootPassword` which set the globals `unomi_root_password`.

But I saw that I forgot the add parameter for `updateHazelcast` action on calls, so I push a new commit for that.

kevinvandel: Oh OK, thanks !
I still can't get used to working with globals to "return" data from an action...

kevinvandel (approved):

PR #8, opened 2021-04-12
PAAS-1325: allow Workato & IT OVH firewall IP addresses

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1325

Short description:

### DIFF
--- a/packages/jcustomer/install.yml
+++ b/packages/jcustomer/install.yml
@@ -13,8 +13,16 @@ description:
     privacy rules (such as GDPR).
 
 globals:
-  jcustomer_env_version: 6
+  jcustomer_env_version: 7
   unomi_version: ${settings.productVersion}
+  stack_connect_ip_adds:
+    - 3.65.225.246
+    - 3.66.45.94
+    - 18.198.249.58
+    - 34.226.132.221
+    - 52.5.142.59
+    - 52.54.43.157
+    - 188.165.59.149
 ssl: true
 skipNodeEmails: true
 
@@ -100,6 +108,7 @@ onBeforeInstall: |
       UNOMI_HTTP_PORT: '80',
       UNOMI_ELASTICSEARCH_ADDRESSES: 'es:' + esPort,
       UNOMI_ELASTICSEARCH_CLUSTERNAME: 'jahia-dx',
+      UNOMI_THIRDPARTY_PROVIDER1_IPADDRESSES: "127.0.0.1,::1,${globals.stack_connect_ip_adds.join(,)}",
       _ROLE: 'unomi',
       _PROVIDE: 'unomi',
       ES_VERSION: esVersion,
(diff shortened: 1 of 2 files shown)


packages/jcustomer/migrations/migrate-to-v4.yaml
```diff
@@ -0,0 +1,74 @@
...
+globals:
+  version: 4
+  workato_ip_adds: "52.5.142.59,34.226.132.221,52.54.43.157,34.204.129.29,34.228.172.35,54.83.143.113,34.192.94.13,34.195.128.7,34.226.84.130"
+  it_ovh_firewall_add: "188.165.59.149"
+
+####
+# Ansible has a weird behavior when receiving a variable in its parameters,
+# it replaces " with ' which leads to an invalid json. To avoid the bug pass the parameters package_settings as follow :
+# -e "package_settings=' {\"papi_hostname\":\"papi[.dev].trial.jahia.com\",\"papi_token\":\"TMP_TOKEN\", \"status\": \"{{status }}\"}'"
+####
```
Ronan-Gorain, line 22: Useless as there is no need for parameters here

packages/jcustomer/install.yml
```diff
@@ -16,6 +16,8 @@ globals:
   jcustomer_env_version: 3
```
kevinvandel, line 16: Since you created the "migrate-to-v4" package, you also need to increment the `jcustomer_env_version` here.

kevinvandel (dismissed):

packages/jcustomer/migrations/migrate-to-v4.yaml
```diff
@@ -0,0 +1,74 @@
...
+        container_env_vars = jelastic.env.control.GetContainerEnvVars(
+                                     "${env.envName}",
+                                     session,
+                                     ${nodes.cp.first.id}
+                                   );
+        allowed_ips = container_env_vars.object.UNOMI_THIRDPARTY_PROVIDER1_IPADDRESSES;
+        if (allowed_ips != "") {
+          allowed_ips = "," + allowed_ips
+        }
+        return {"result": 0, "UNOMI_THIRDPARTY_PROVIDER1_IPADDRESSES": "127.0.0.1,::1,${globals.workato_ip_adds},${globals.it_ovh_firewall_add}," + allowed_ips}
```
Ronan-Gorain, line 70: Unless I'm mistaken, the env var still needs to be updated. Also, maybe you just have to call the api in the script, no need to return it.

Ronan-Gorain (changes requested):

packages/jcustomer/migrations/migrate-to-v4.yaml
adoux: Very good catch, forgot the essential part :facepalm: 
Fixed

packages/jcustomer/migrations/migrate-to-v4.yaml
adoux: Removed!

packages/jcustomer/install.yml
adoux: Good catch, fixed!

laurentfufu (approved):

PR #9, opened 2021-04-16
Paas 1541

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-1541

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v5.yaml
+++ b/packages/jahia/migrations/migrate-to-v5.yaml
@@ -1,0 +1,54 @@
+---
+type: update
+version: 1.5.2
+name: Migrate Jahia env to v5
+id: migrate-jahia-env-v5
+
+# Required for healthchecks
+mixins:
+  - "../../../mixins/common.yml"
+  - "../../../mixins/jahia.yml"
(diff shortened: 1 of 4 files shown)


packages/jahia/migrations/migrate-to-v5.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+# Required for healthchecks
+mixins:
+  - "../../../mixins/common.yml"
+  - "../../../mixins/jahia.yml"
+  - "../../../mixins/mariadb.yml"
+  - "../../../mixins/haproxy.yml"
+
+globals:
+  version: 5
```
kevinvandel, line 15: From now on, when creating a new migration package, it's also needed to increment the env_version number in the install package (otherwise new envs will have a wrong version).

kevinvandel (changes requested):

packages/jahia/migrations/migrate-to-v5.yaml
Ronan-Gorain: Ok, it's fixed

kevinvandel (approved):

amitc-jahia (approved):

PR #12, opened 2021-04-21
PAAS-1535: fix karaf password when containing dollar sign

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1535

Short description:
I tried to think of a simple way to do it but it was more complex than it seemed because of the `export` command in `setenv` file.
The way I choose to go with allow the user to put single/double quotes, dollar signs and other special characters in their jcustomer password.
I've tested some scenarios and it seems to work pretty well

### DIFF
--- a/assets/jcustomer/reset-unomi-root-password.py
+++ b/assets/jcustomer/reset-unomi-root-password.py
@@ -6,21 +6,22 @@ from binascii import a2b_base64
 import re
 import os.path
 
+new_password_b64 = sys.argv[1]
 new_password = a2b_base64(str.encode(sys.argv[1])).decode("utf-8")
 unomi_env_file = sys.argv[2]
 datadog_conf_file = sys.argv[3]
 
 pwd_set = False
-password_line = "export UNOMI_ROOT_PASSWORD=" + new_password
+password_line = f'UNOMI_ROOT_PASSWORD_B64="{new_password_b64}"\n'
+export_password_line = "export UNOMI_ROOT_PASSWORD=$(echo $UNOMI_ROOT_PASSWORD_B64 | base64 -d)"
 with fileinput.FileInput(unomi_env_file, inplace=True) as file:
     for line in file:
-        if "UNOMI_ROOT_PASSWORD" in line:
-            line = password_line
-            pwd_set = True
+        if "UNOMI_ROOT_PASSWORD_B64=" in line or "export UNOMI_ROOT_PASSWORD=" in line:
+            continue
         print(line, end='')
-if not pwd_set:
-    with open(unomi_env_file, "a") as file:
-        file.write(password_line)
+
+with open(unomi_env_file, 'a') as file:
+    file.write(password_line + export_password_line)
 
 if not os.path.exists(datadog_conf_file):
     exit(0)
--- a/assets/jcustomer/reset-unomi-root-password.py
+++ b/assets/jcustomer/reset-unomi-root-password.py
@@ -13,7 +13,7 @@ datadog_conf_file = sys.argv[3]
 
 pwd_set = False
 password_line = f'UNOMI_ROOT_PASSWORD_B64="{new_password_b64}"\n'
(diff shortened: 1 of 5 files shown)


amitc-jahia (commented):
For the _reset-unomi-root-password.py_ script, I don't think we need to make any changes as it is already able to take special characters and the password doesn't get truncated. Also, in the case of password reset, the new changes would only update **UNOMI_ROOT_PASSWORD_B64** and not the **UNOMI_ROOT_PASSWORD** in the envfile.

assets/jcustomer/reset-unomi-root-password.py
```diff
@@ -6,18 +6,19 @@
...
-password_line = "export UNOMI_ROOT_PASSWORD=" + new_password
+password_line = f'UNOMI_ROOT_PASSWORD_B64="{new_password_b64}"'
 with fileinput.FileInput(unomi_env_file, inplace=True) as file:
     for line in file:
-        if "UNOMI_ROOT_PASSWORD" in line:
+        if "UNOMI_ROOT_PASSWORD_B64=" in line:
             line = password_line
             pwd_set = True
-        print(line, end='')
+        print(line)
```
laurentfufu, line 21: I guess this will add ugly empty lines every time it's launched

assets/jcustomer/reset-unomi-root-password.py
```diff
@@ -6,18 +6,19 @@
...
 unomi_env_file = sys.argv[2]
 datadog_conf_file = sys.argv[3]
 
 pwd_set = False
-password_line = "export UNOMI_ROOT_PASSWORD=" + new_password
+password_line = f'UNOMI_ROOT_PASSWORD_B64="{new_password_b64}"'
 with fileinput.FileInput(unomi_env_file, inplace=True) as file:
     for line in file:
-        if "UNOMI_ROOT_PASSWORD" in line:
+        if "UNOMI_ROOT_PASSWORD_B64=" in line:
```
laurentfufu, line 18: what about setenv without this line ?

laurentfufu (changes requested):

assets/jcustomer/reset-unomi-root-password.py
adoux: Indeed, fixed!

adoux:
> For the _reset-unomi-root-password.py_ script, I don't think we need to make any changes as it is already able to take special characters and the password doesn't get truncated. Also, in the case of password reset, the new changes would only update **UNOMI_ROOT_PASSWORD_B64** and not the **UNOMI_ROOT_PASSWORD** in the envfile.

Yeah but this is actually what we want, the `UNOMI_ROOT_PASSWORD` will be updated as it decodes UNOMI_ROOT_PASSWORD_B64 before exporting the variable.
We have to do it that way if we want to allow special characters like dollar sign, equal sign or quotes.

assets/jcustomer/reset-unomi-root-password.py
adoux: I actually forgot the migration package so you're right :man_facepalming: 
I've updated the script anyways so it will always add the two required lines at the end of the file, and I added the migration package as well

laurentfufu (approved):

assets/jcustomer/reset-unomi-root-password.py
```diff
@@ -6,21 +6,22 @@
...
 
+new_password_b64 = sys.argv[1]
 new_password = a2b_base64(str.encode(sys.argv[1])).decode("utf-8")
 unomi_env_file = sys.argv[2]
 datadog_conf_file = sys.argv[3]
 
 pwd_set = False
-password_line = "export UNOMI_ROOT_PASSWORD=" + new_password
+password_line = f'UNOMI_ROOT_PASSWORD_B64="{new_password_b64}"\n'
+export_password_line = "export UNOMI_ROOT_PASSWORD=$(echo $UNOMI_ROOT_PASSWORD_B64 | base64 -d)"
```
laurentfufu, line 16: I think a newline would have been nice here

adoux: Done!

laurentfufu (approved):

PR #26, opened 2021-05-07
PAAS-1595: fix jcustomer random password empty

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1595

Short description:

### DIFF
--- a/packages/jcustomer/install.yml
+++ b/packages/jcustomer/install.yml
@@ -15,7 +15,6 @@ description:
 globals:
   jcustomer_env_version: 4
   unomi_version: ${settings.productVersion}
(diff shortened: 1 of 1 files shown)


packages/jcustomer/install.yml
```diff
@@ -99,6 +99,8 @@ nodes: definedInOnBeforeInstall
 
 # --Events --
 onInstall:
+  - setGlobals:
+      unomi_root_password_b64: ${globals.random_unomi_root_password.toBase64()}
```
kevinvandel, line 103: Why not just use an `else` clause after the `if(settings.rootPassword)` ?

adoux: Done, force pushed

PR #28, opened 2021-05-10
PAAS-1557: Use non-default apiKey for jCustomer provider

amitc-jahia:
JIRA issue: https://jira.jahia.org/browse/PAAS-1557

Short description:

### DIFF
--- a/mixins/jcustomer.yml
+++ b/mixins/jcustomer.yml
@@ -257,6 +257,14 @@ actions:
         openssl rand -hex 256 | sha1sum | cut -d' ' -f1
     - setGlobals:
         apiKey: ${response.out}
-    - api[${this}]: env.control.AddContainerEnvVars
+    - api[cp]: env.control.AddContainerEnvVars
       vars: {"UNOMI_THIRDPARTY_PROVIDER1_KEY": "${globals.apiKey}"}
     - cmd[cp]: systemctl restart karaf
+
+  getJcustomerKey:
+    - script: |
+        const jcustomerEnv = "${this.jcustomerEnv}"
+        resp = jelastic.env.control.GetContainerEnvVarsByGroup(jcustomerEnv, session, nodeGroup='cp');
+        return {'result': 0, 'out': resp.object["UNOMI_THIRDPARTY_PROVIDER1_KEY"]};
+    - setGlobals:
+        jcustomerKey: ${response.out}
--- a/mixins/jcustomer.yml
+++ b/mixins/jcustomer.yml
@@ -251,3 +251,13 @@ actions:
           }
         );
         return {"result": 0, "out": "Jahia environments updated " + res};
+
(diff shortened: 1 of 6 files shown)


packages/jahia/link-to-jcustomer.yml
```diff
@@ -143,6 +144,13 @@ actions:
             type: error
             message: "An error occurred during jexperience installation."
 
+  getJcustomerKey:
```
Ronan-Gorain, line 147: This should be placed into the mixin file like the setJcustomerKey action. So it could be re-used in migrate-to-v6-package

mixins/jcustomer.yml
```diff
@@ -251,3 +251,12 @@ actions:
           }
         );
         return {"result": 0, "out": "Jahia environments updated " + res};
+
+  setJcustomerApiKey:
+    - cmd[${nodes.cp.first.id}]: |-
+        openssl rand -hex 256 | sha1sum | cut -d' ' -f1
+    - setGlobals:
+        apiKey: ${response.out}
+    - api[${this}]: env.control.AddContainerEnvVars
```
Ronan-Gorain, line 260: You are using "cp" nodes explicitly everywhere else in this action. Maybe the "cp" argument used to call the action is useless and you should use api[cp] here. 
Otherwise, if there is a advantage to use a parameter, use this ${this} on others cmd blocks :)

packages/jcustomer/migrations/migrate-to-v5.yaml
```diff
@@ -0,0 +1,46 @@
...
+      text: "Environment $envName migrated to v${globals.version}"
+  - checkJcustomerHealth: cp
+  - checkEsClusterStatus
+
+actions:
+  eventsUpdate:
+    install:
+      jps: "${baseUrl}/../update-events.yml"
+
+  updateJcustomerApiKey:
```
Ronan-Gorain, line 39: Use the setJcustomerApiKey method defined in mixin.
Moreover, you are using a ${this} although there is no parameter provided when you call this action.

Ronan-Gorain (dismissed):

mixins/jcustomer.yml
amitc-jahia: Fixed

packages/jcustomer/migrations/migrate-to-v5.yaml
amitc-jahia: Fixed

mixins/jcustomer.yml
amitc-jahia: Fixed

packages/jahia/link-to-jcustomer.yml
amitc-jahia: Action moved to mixin

adoux (approved):

PR #43, opened 2021-05-31
PAAS-1592 add dns split for ipsec in migration script

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1592

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v6.yaml
+++ b/packages/jahia/migrations/migrate-to-v6.yaml
@@ -31,6 +31,7 @@ onInstall:
   - removeStrongSwanServiceFromRedeploy  # PAAS-1587
   - datadogCustomCheckForJahiaNodes      # PAAS-1596
   - setJahiaTcpBindAddress               # PAAS-1580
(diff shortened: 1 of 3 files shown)


packages/jahia/migrations/migrate-to-v6.yaml
```diff
@@ -132,3 +133,15 @@ actions:
       if [ "$(grep jahia_cfg_cluster_tcp_bindAddress /opt/tomcat/conf/tomcat-env.sh)" == "" ]; then
         sed -i '/TOMCAT_USER=/ a jahia_cfg_cluster_tcp_bindAddress=$(hostname -i)' /opt/tomcat/conf/tomcat-env.sh
       fi
+
+  installIPSecDNSSplit:
```
adoux, line 137: Shouldn't we make this action idempotent? For instance testing if dnsmasq is installed before doing anything.
I think we should at least make the `sed` idempotent.

laurentfufu: done

adoux (approved):

PR #45, opened 2021-06-02
PAAS-1590: Check if righsubnets don't overlap with env subnet in IPse…

amitc-jahia:
…c package

JIRA issue: https://jira.jahia.org/browse/PAAS-1590

Short description:

### DIFF
--- a/packages/jahia/ipsec.yml
+++ b/packages/jahia/ipsec.yml
@@ -86,6 +86,27 @@ actions:
           exit 1
         fi
 
+        # check if right subnets overlap with env's subnet
+        pip install ipconflict
+
+        infra_subnet=192.168.0.0/16
+        subnets_file=$(mktemp)
+
+        cat > $subnets_file << END
+        $infra_subnet
+        END
+        
+        grep 'rightsubnet' $conf_tmpfile | grep -v '#' | grep -Eo "([0-9]{1,3}\.){3}[0-9]{1,3}(\/[0-9]{1,})?" | uniq >> $subnets_file
+        env_ip=$(hostname -i)
+
+        while read subnet; do
+          ipconflict -q $subnet $env_ip
+          if [ $? -eq 0 ]; then
+            echo "[ERROR] Rightsubnet overlaps with Env subnet" >&2
+            exit 1
+          fi
+        done < $subnets_file
+        
         # work on connection conf
         new_conf_md5=$(md5sum $conf_tmpfile | cut -d" " -f1)
         if [ -f $confdir/$confname.conf ]; then


packages/jahia/ipsec.yml
```diff
@@ -86,6 +86,25 @@ actions:
...
+        pip install ipconflict
+
+        infra_subnet=192.168.0.0/16
+        subnets_file=$(mktemp)
+
+        cat > $subnets_file << END
+        $infra_subnet
+        END
+
+        cat $conf_tmpfile | grep rightsubnet | cut -d'=' -f2 | tr -d ' ' | uniq >> $subnets_file
```
laurentfufu, line 99: 2 things bother me here:
* possible commented `rightsubnet` line will be taken into account
* `rightsubnet` can be in the form `<ip subnet>[[<proto/port>]][,...]` so this will work only with one entry (ip or subnet notation)

laurentfufu (changes requested):

packages/jahia/ipsec.yml
amitc-jahia: Fixed

packages/jahia/ipsec.yml
```diff
@@ -86,6 +86,25 @@ actions:
...
+        pip install ipconflict
+
+        infra_subnet=192.168.0.0/16
+        subnets_file=$(mktemp)
+
+        cat > $subnets_file << END
+        $infra_subnet
+        END
+        
+        grep 'rightsubnet' $conf_tmpfile | grep -v '#' | grep -o "[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\/[0-9]\{1,\}" | uniq >> $subnets_file
```
laurentfufu, line 99:
```suggestion
        grep 'rightsubnet' $conf_tmpfile | grep -v '#' | grep -Eo "([0-9]{1,3}\.){3}[0-9]{1,3}(\/[0-9]{1,})?" | uniq >> $subnets_file
```
imo, grouping the first 3 bytes and using grep's switch `-E` in order to not have to backslash brackets will help readability for this regex. As the network mask bits is optional (strongswan will assume it's `/32` then), I also update this part of the regex.

packages/jahia/ipsec.yml
```diff
@@ -86,6 +86,25 @@ actions:
...
+        env_ip=$(hostname -i)
+
+        ipconflict -qf $subnets_file $env_ip
```
laurentfufu, line 102: Not sure about testing all `rightsubnet` in one time because each entry will be tested against each other and not only against the jelastic/env's subnets. Customer may be wrong itself add why not having something like `10.100.0.0/16,10.100.2.0/24` and this will trigger an error telling that this "overlaps with Env subnet".

laurentfufu (changes requested):

packages/jahia/ipsec.yml
amitc-jahia: Fixed

packages/jahia/ipsec.yml
amitc-jahia: Fixed

laurentfufu (approved):

PR #52, opened 2021-06-15
PAAS-1609: Rework all version comparisons to use 'compareVersions' ac…

amitc-jahia:
…tion

JIRA issue: https://jira.jahia.org/browse/PAAS-1609

Short description:

### DIFF
--- a/mixins/common.yml
+++ b/mixins/common.yml
@@ -126,7 +126,7 @@ actions:
         versionIsLower: -1
         ${this.res}: ${response.outcome}
 
-  isVersionLower:
+  isVersionStrictlyLower:
     # Parameters :
     #  a : the first version to compare
     #  b : the second version to compare
@@ -137,15 +137,15 @@ actions:
     - compareVersions:
         a: "${this.a}"
         b: "${this.b}"
-        res: isVersionLowerComparison
-    - if (${globals.isVersionLowerComparison} == -1):
+        res: isVersionStrictlyLowerComparison
+    - if (${globals.isVersionStrictlyLowerComparison} == -1):
         setGlobals:
           ${this.res}: true
     - else:
         setGlobals:
           ${this.res}: false
 
-  isVersionHigher:
+  isVersionStrictlyHigher:
     # Parameters :
     #  a : the first version to compare
     #  b : the second version to compare
@@ -156,8 +156,8 @@ actions:
     - compareVersions:
         a: "${this.a}"
         b: "${this.b}"
-        res: isVersionHigherComparison
-    - if (${globals.isVersionHigherComparison} == 1):
+        res: isVersionStrictlyHigherComparison
+    - if (${globals.isVersionStrictlyHigherComparison} == 1):
         setGlobals:
           ${this.res}: true
     - else:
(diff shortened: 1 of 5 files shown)


packages/common/restore.yml
```diff
@@ -109,13 +109,26 @@ actions:
...
+        a: ${globals.jahiaVersion}
+        b: 7.2.3.2
+        res: isVersionLowerThan7232
+    - if ( ${globals.isVersionLowerThan7232}):
+        - cmd [proc,cp]: touch /data/digital-factory-data/backup-restore
+    - else:
+        - cmd [proc,cp]: |-
+            touch /data/digital-factory-data/safe-env-clone
+            short_name=$(echo ${_ROLE}.$HOSTNAME | sed -r 's/^([a-Z]+)\.[a-Z]+([0-9]+)-.+$/\1.\2/' | tr 'A-Z' 'a-z')
+            sed -i "s|^cluster.node.serverId.*|cluster.node.serverId = $short_name|" $STACK_PATH/conf/digital-factory-config/jahia/jahia.node.properties
```
laurentfufu, line 127: shouldn't this be done whatever the version of jahia ?

packages/jahia/perf-test-step1.yml
```diff
@@ -11,9 +11,21 @@ globals:
...
+      a: ${globals.jahiaVersion}
+      b: 8.0.0.0
+      res: isVersionHigherThan8000
+
+  - isVersionEqual:
+      a: ${globals.jahiaVersion}
+      b: 8.0.0.0
+      res: isVersionEqualTo8000
+
+  - if( ${globals.isVersionHigherThan8000} || ${globals.isVersionEqualTo8000}):
```
laurentfufu, line 28: maybe we can do something like that here (or create `isVersionHigherOrEqual` and `isVersionLowerOrEqual`, sure they will be useful)

```suggestion
  - Compare:
      a: ${globals.jahiaVersion}
      b: 8.0.0.0
      res: isVersionHigherOrEqual8000

  - if(globals.isVersionHigherOrEqual8000 >= 0):
```

packages/common/restore.yml
kevinvandel: Yes indeed, we should completely remove this condition since we don't have any version older than 7.3
It seems we forgot to delete it before.

packages/jahia/perf-test-step1.yml
kevinvandel: Personally I think that creating `isVersionHigherOrEqual` and `isVersionLowerOrEqual` (and also renaming the others 2 methods into `isVersionStrictlyHigher` and `isVersionStriclyLower`) would help make things easier to understand ; btw that's what we did on the frontend side. When I started my review I had to check the action definition to see if the condition was strict or not...

An alternative in your case @amitc-jahia would have been to check this instead:
```
  - isVersionLower:
      a: ${globals.jahiaVersion}
      b: 8.0.0.0
      res: isVersionLowerThan8000
  - if( ! ${globals.isVersionLowerThan8000} ):
```
But this isn't the easiest to understand, having multiple actions with explicit names is the way to go IMHO.

kevinvandel (changes requested):

packages/common/restore.yml
amitc-jahia: Fixed

packages/jahia/perf-test-step1.yml
amitc-jahia: Fixed

mixins/common.yml
```diff
@@ -240,6 +240,44 @@ actions:
...
+    #  b : the second version to compare
+    #  res : the global variable name to store the result
+    # Returns:
+    # 1 if a <= b
+    # 0 if a > b
+    - compareVersions:
+        a: "${this.a}"
+        b: "${this.b}"
+        res: isVersionLowerOrEqualComparison
+    - if (${globals.isVersionLowerOrEqualComparison} == -1 || ${globals.isVersionLowerOrEqualComparison} == 0):
```
kevinvandel, line 255:
```suggestion
    - if (${globals.isVersionLowerOrEqualComparison} <= 0):
```

mixins/common.yml
```diff
@@ -240,6 +240,44 @@ actions:
...
+    #  b : the second version to compare
+    #  res : the global variable name to store the result
+    # Returns:
+    # 1 if a >= b
+    # 0 if a < b
+    - compareVersions:
+        a: "${this.a}"
+        b: "${this.b}"
+        res: isVersionHigherOrEqualComparison
+    - if (${globals.isVersionLowerOrEqualComparison} == 1 || ${globals.isVersionLowerOrEqualComparison} == 0):
```
kevinvandel, line 274:
```suggestion
    - if (${globals.isVersionLowerOrEqualComparison} >= 0):
```

kevinvandel (changes requested):

mixins/common.yml
amitc-jahia: FIxed

mixins/common.yml
amitc-jahia: FIxed

kevinvandel (approved):

PR #53, opened 2021-06-15
PAAS-1578  Set more reliable values for UNOMI_CLUSTER_*_ADDRESS

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1578

Short description:


kevinvandel (commented):
@laurentfufu no migration needed to update current environments?

kevinvandel (approved):

PR #57, opened 2021-06-17
PAAS-1643: Package to fix 'context-patch' index configuration

amitc-jahia:
JIRA issue: https://jira.jahia.org/browse/PAAS-1643

Short description:

### DIFF
--- a/packages/one-shot/fix-context-patch-index-configuraton.yml
+++ b/packages/one-shot/fix-context-patch-index-configuraton.yml
@@ -1,0 +1,63 @@
+---
+type: update
+version: 1.5.2
+name: jCustomer - Fix context-patch index configuration on existing environments
+id: jcustomer-fix-context-patch-index-configuration
+description:
+  short: jCustomer - Fix context-patch index configuration on existing environments
+
+onInstall:
+  - cmd[cp]: |-
+      curl -s $UNOMI_ELASTICSEARCH_ADDRESSES/_cat/indices | awk '{print $3}' | grep 'dx_*'
+      if [ $? -eq 0 ]; then
+        echo "[INFO] This is an Augmented Search env. Nothing to do here" >&2
+        exit 1
+      fi
+      curl -s $UNOMI_ELASTICSEARCH_ADDRESSES/_cat/indices | awk '{print $3}' | grep 'context-patch'
+      if [ $? -ne 0 ]; then
+        echo "[ERROR] Index 'context-patch' not present" >&2
+        exit 2
+      fi
+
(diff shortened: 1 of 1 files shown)


packages/one-shot/fix-context-patch-index-configuraton.yml
```diff
@@ -0,0 +1,63 @@
...
+type: update
+version: 1.5.2
+name: jCustomer - Fix context-patch index configuration on existing environments
+id: jcustomer-fix-context-patch-index-configuration
+description:
+  short: jCustomer - Fix context-patch index configuration on existing environments
+
+onInstall:
+  - cmd[cp]: |-
+      curl -s $UNOMI_ELASTICSEARCH_ADDRESSES/_cat/indices | awk '{print $3}' | grep 'dx_*'
```
adoux, line 11: You can use filters directly in the request, like `es:9200/_cat/indices?h=index`, that way you don't need to `awk`

(Same thing for the next curls)

packages/one-shot/fix-context-patch-index-configuraton.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      # Create a new temp index
+      curl -s -XPUT $UNOMI_ELASTICSEARCH_ADDRESSES/context-patch-temp -H 'Content-Type: application/json' \
+        -d '{"settings": {"number_of_shards": '"$desired_shards"',"number_of_replicas": '"$desired_replicas"'}}'
+
+      # Reindex context-patch to temp index
+      curl -s -XPOST $UNOMI_ELASTICSEARCH_ADDRESSES/_reindex -H 'Content-Type: application/json' \
+        -d '{"source": {"index": "context-patch"},"dest": {"index": "context-patch-temp"}}'
+
+      curl -s -XDELETE $UNOMI_ELASTICSEARCH_ADDRESSES/context-patch
```
adoux, line 40: I think we should make sure that the previous command is OK before deleting the index, just in case.

Overall I think you can add something like `| exit 1` after curl POST/PUT commands since they all depend on the previous ones

adoux (changes requested):

packages/one-shot/fix-context-patch-index-configuraton.yml
amitc-jahia: Done

packages/one-shot/fix-context-patch-index-configuraton.yml
amitc-jahia: Changes made

adoux (approved):

PR #60, opened 2021-06-18
PAAS-1578 update envvars only, split restart to another action

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1578

Short description:

### DIFF
--- a/packages/jcustomer/migrations/migrate-to-v6.yaml
+++ b/packages/jcustomer/migrations/migrate-to-v6.yaml
@@ -21,10 +21,10 @@ onInstall:
   - eventsUpdate
 
   - updateClusterAddresses                        # PAAS-1578
+  - rollingRestartJc
 
   ### Post-actions
   - setEnvVersion: ${globals.version}
-  - rollingRestartJc
   - logEvent:
       target: ${nodes.cp.first.id}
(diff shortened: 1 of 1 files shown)


packages/jcustomer/migrations/migrate-to-v6.yaml
```diff
@@ -20,10 +20,11 @@ onInstall:
   - checkEsClusterStatus
   - eventsUpdate
 
-  - updateHazelcast                               # PAAS-1578
+  - updateClusterAddresses                        # PAAS-1578
 
   ### Post-actions
   - setEnvVersion: ${globals.version}
+  - rollingRestartJc
```
Ronan-Gorain, line 27: It would be better to put it in at the end of the previous section. This one is supposed to be always the same and someone could miss the line.

Ronan-Gorain (changes requested):

packages/jcustomer/migrations/migrate-to-v6.yaml
laurentfufu: resurrected :trollface:

Ronan-Gorain (approved):

PR #70, opened 2021-06-23
PAAS-1653: Fix custom check script for jahia_node_not_in_haproxy_pool monitor

amitc-jahia:
JIRA issue: https://jira.jahia.org/browse/PAAS-1653

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v7.yaml
+++ b/packages/jahia/migrations/migrate-to-v7.yaml
@@ -25,13 +25,10 @@ onInstall:
   - eventsUpdate
   ### End Pre-checks
 
-<<<<<<< HEAD
   - updateIPSecScript         # PAAS-1651
   - updateDnsmasqConf         # PAAS-1651
-=======
   - updateHaproxyHealthcheckRequests              # PAAS-1653
-  - changeAccessLogPatternForJahiaNode            # PAAS-1653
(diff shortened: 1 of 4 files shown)


packages/jahia/migrations/migrate-to-v7.yaml
```diff
@@ -0,0 +1,99 @@
...
+  - checkEnvVersion: ${globals.version}
+  - checkJahiaHealth: "cp, proc"
+  - checkHaproxyHealth: bl
+  - checkMariadbHealth: sqldb
+  - checkGaleraClusterHealth: sqldb
+  - eventsUpdate
+  ### End Pre-checks
+
+  - updateHaproxyHealthcheckRequests              # PAAS-1653
+  - changeAccessLogPatternForJahiaNode            # PAAS-1653
```
kevinvandel, line 29: We should add a BIG warning here that `changeAccessLogPatternForJahiaNode` will restart all jahia nodes. Or split the restart in a specific step

kevinvandel (changes requested):

packages/jahia/migrations/migrate-to-v7.yaml
amitc-jahia: Update pattern and restart jahia steps have been split into 2 actions.

kevinvandel (approved):

PR #73, opened 2021-06-29
PAAS-1645 install ImageMagick7 from remi's repo instead of AppImage

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1645

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v7.yaml
+++ b/packages/jahia/migrations/migrate-to-v7.yaml
@@ -29,6 +29,7 @@ onInstall:
   - updateDnsmasqConf                 # PAAS-1651
   - addJelasticExposeFalse            # PAAS-1658
   - updateStrongswanOverride          # PAAS-1667
+  - updateIMFlavor                    # PAAS-1645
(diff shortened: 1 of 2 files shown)


packages/jahia/migrations/migrate-to-v7.yaml
```diff
@@ -138,3 +139,13 @@ actions:
...
     - env.control.AddContainerEnvVars[sqldb]:
       vars: {"JELASTIC_EXPOSE": "false"}
+
+  updateIMFlavor:
+    - cmd[cp, proc]: |-
+        yum install -y https://rpms.remirepo.net/enterprise/remi-release-7.rpm
+        yum --enablerepo=remi install -y ImageMagick7
+        [ -f /usr/bin/magick ] || exit 1
+      user: root
+    - env.control.AddContainerEnvVars [cp, proc]:
```
amitc-jahia, line 149: Can add an idempotent check if it is already installed
```suggestion
    -  if ("${response.errOut}" == ""):
          env.control.AddContainerEnvVars [cp, proc]:
```

amitc-jahia (approved):

PR #84, opened 2021-07-07
Paas 1321 rethink haproxy conf

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1321

Short description:

### DIFF
--- a/packages/common/restore.yml
+++ b/packages/common/restore.yml
@@ -299,9 +299,8 @@ actions:
         cluster_role=$(awk -F'=' '$1=="JEL_ENV_ROLE" {print $2}' /metadata_from_HOST); export cluster_role
         env_region=$(awk -F'=' '$1=="JEL_REGION" {print $2}' /metadata_from_HOST | sed 's/[[:punct:]]//g'); export env_region
         cd jelastic_backup
-        if python3 backrest.py -a download --backupname ${settings.backup_name} --bucketname ${globals.bucketname} -m ${settings.backtype} -t ${settings.timestamp} -f haproxy.tar.gz 2>>${globals.logsPath}; then
-          ls haproxy.tar.gz || exit 0
-        fi
+        python3 backrest.py -a download --backupname ${settings.backup_name} --bucketname ${globals.bucketname} -m ${settings.backtype} -t ${settings.timestamp} -f haproxy.tar.gz 2>>${globals.logsPath}
+        [ -f haproxy.tar.gz ] && exit 0
         python3 backrest.py -a download --backupname ${settings.backup_name} --bucketname ${globals.bucketname} -m ${settings.backtype} -t ${settings.timestamp} -f haproxy-00-global.cfg.gz 2>>${globals.logsPath} || exit 1
         ls haproxy-00-global.cfg.gz || exit 0
     - if ("${response.errOut}" == ""):
--- a/packages/common/restore.yml
+++ b/packages/common/restore.yml
@@ -315,6 +315,7 @@ actions:
         AUTH_BASIC=auth_basic
         CUSTOMER_RULES_FILE=customer_rules.cfg
         HAPROXY_CONF_DIR=/etc/haproxy/haproxy.cfg.jahia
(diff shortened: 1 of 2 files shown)


packages/common/restore.yml
```diff
@@ -299,10 +299,9 @@ actions:
...
         env_region=$(awk -F'=' '$1=="JEL_REGION" {print $2}' /metadata_from_HOST | sed 's/[[:punct:]]//g'); export env_region
         cd jelastic_backup
-        if python3 backrest.py -a download --backupname ${settings.backup_name} --bucketname ${globals.bucketname} -m ${settings.backtype} -t ${settings.timestamp} -f haproxy.tar.gz 2>>${globals.logsPath}; then
-          ls haproxy.tar.gz || exit 0
-        fi
-        python3 backrest.py -a download --backupname ${settings.backup_name} --bucketname ${globals.bucketname} -m ${settings.backtype} -t ${settings.timestamp} -f haproxy-00-global.cfg.gz 2>>${globals.logsPath} || exit 1
+        python3 backrest.py -a download --backupname ${settings.backup_name} --bucketname ${globals.bucketname} -m ${settings.backtype} -t ${settings.timestamp} -f haproxy.tar.gz 2>>${globals.logsPath}
+        [ -f haproxy.tar.gz ] && exit 0
+        python3 backrest.py -a download --backupname ${settings.backup_name} --bucketname ${globals.bucketname} -m ${settings.backtype} -t ${settings.timestamp} -f haproxy-00-global.cfg.gz 2>>${globals.logsPath}
```
Ronan-Gorain, line 304: I think the || exit 1 was mandatory here to support (very, very) old backups that have no haproxy conf. It's been so long that we could get rid of this backward compatibility, but then we have rework the next if so to exit with error if errOut is not  empty

packages/jahia/migrations/migrate-to-v7.yaml
```diff
@@ -178,31 +178,21 @@ actions:
...
-          echo $(awk '/user .* password .*/ {print $2" "$4}' $OLD_HAPROXY_CONF)
+        HAPROXY_DIR=/etc/haproxy
+        OLD_HAPROXY_CONF=$HAPROXY_DIR/haproxy.cfg.d/00-global.cfg
+        HAPROXY_CONF=$HAPROXY_DIR/haproxy.cfg.jahia/jahia-cloud.cfg
+        user=$(awk '/user .* password .*/ {print $2}' $OLD_HAPROXY_CONF)
+        password=$(awk '/user .* password .*/ {print $4}' $OLD_HAPROXY_CONF)
+        sed -i "s;^\(\s*user \).*\( password \).*$;\1$user\2$password;" $HAPROXY_CONF
+        old_auth_basic_disabled=$(grep -c  "#acl tools.*#HTTP_AUTH_BASIC" $OLD_HAPROXY_CONF)
+        auth_basic_disabled=$(grep -c  "#acl tools.*#HTTP_AUTH_BASIC" $HAPROXY_CONF)
+        [ $old_auth_basic_disabled -eq $auth_basic_disabled ] && echo OUI
```
Ronan-Gorain, line 189: && echo OUI ? :)

Ronan-Gorain (changes requested):

packages/common/restore.yml
kevinvandel: > It's been so long that we could get rid of this backward compatibility

I think we can get rid of it indeed.

packages/jahia/migrations/migrate-to-v7.yaml
adoux: :facepalm: :facepalm: 
Fixed

packages/common/restore.yml
adoux: I added the `|| exit 1` again

kevinvandel (approved):

packages/jahia/migrations/migrate-to-v7.yaml
kevinvandel: echo NON :trollface:

PR #96, opened 2021-07-26
PAAS-1631: Add organization name in env vars

amitc-jahia:
JIRA issue: https://jira.jahia.org/browse/PAAS-1631

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v8.yaml
+++ b/packages/jahia/migrations/migrate-to-v8.yaml
@@ -26,11 +26,12 @@ onInstall:
   - setGlobalRepoRootUrl
   ### End Pre-checks
 
-  - addHaproxyNodesCountOnJahiaNodes  # PAAS-1674
-  - enableGCMetricsOnJahiaNodes       # PAAS-1664
-  - fixTomcatInitScript               # PAAS-1685
-  - addMaxPostSize                    # PAAS-1579
-  - addGaleraNodeCustomCheck          # PAAS-1677
+  - addHaproxyNodesCountOnJahiaNodes                       # PAAS-1674
+  - enableGCMetricsOnJahiaNodes                            # PAAS-1664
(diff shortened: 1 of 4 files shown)


laurentfufu (commented):
I have the feeling that the implementation of this envvar has more its place in the onBeforeInstall rather than via an action launched after the creation of the nodes. It would be more advantageous in terms of instantiation time.

amitc-jahia:
Changes made. Moved the adding of the envvar to onBeforeInstall rather than a post install action. Migration package updated too.

packages/jahia/migrations/migrate-to-v8.yaml
```diff
@@ -94,3 +95,10 @@ actions:
         chown -h dd-agent: $custom_check_config
         systemctl restart datadog-agent
       user: root
+
+  setOrganizationNameEnvVar:
+    - cmd[${this}]: echo "${user.email}" | cut -d'@' -f1
+    - setGlobals:
+        organization_name: ${response.out}
+    - env.control.AddContainerEnvVars[${this}]:
+      vars: {"ORGANIZATION_NAME": "${globals.organization_name}"}
```
laurentfufu, line 104:
```suggestion

  setOrganizationNameEnvVar:
    - if(!globals.orgname):
      - script: |-
         return {"result": 0, "orgname": "${user.email}".split('@')[0]}
      - setGlobals:
         organization_name: ${response.orgname}
    - env.control.AddContainerEnvVars[${this}]:
      vars: {"ORGANIZATION_NAME": "${globals.organization_name}"}
```


This suggestion should only set the global once (and in a much faster way) instead of reset it on each nodes for each nodegroup.

amitc-jahia: Changes made

laurentfufu (approved):

PR #108, opened 2021-08-05
PAAS-1498: Handle multiple haproxy custom rules files in jelastic pac…

amitc-jahia:
…kage

JIRA issue: https://jira.jahia.org/browse/PAAS-1498

Short description:

### DIFF
--- a/packages/jahia/rewrite-rules.yml
+++ b/packages/jahia/rewrite-rules.yml
@@ -8,10 +8,8 @@ description:
 
 globals:
   rules_dir: "/etc/haproxy/haproxy.cfg.jahia/customer.configuration.d"
-  rules_filename: "customer_rules.cfg"
-  rules_file: "${globals.rules_dir}/${globals.rules_filename}"
-  backup_rules_file: "/var/lib/haproxy/${globals.rules_filename}.bak"
-  temp_rules_file: "/var/lib/haproxy/${globals.rules_filename}.cfg"
+  backup_rules_dir: "/var/lib/haproxy/customer.configuration.d"
+  temp_rules_dir: "/var/lib/haproxy/customer.configuration.tmp"
 
 onInstall:
   - getRewritesRules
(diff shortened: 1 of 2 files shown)


packages/jahia/rewrite-rules.yml
```diff
@@ -24,24 +21,36 @@ actions:
...
 
-        if echo "${settings.md5sum} ${globals.temp_rules_file}" | md5sum --check --status ; then
-          mv -f ${globals.temp_rules_file} ${globals.rules_file}
-        else
-          echo "MD5 sum mismatch. Aborting" >&2
-        fi
+        # Get all rules secret and store as individual files
+        for rules_set in $(echo $rules_sets | jq -r 'keys[]'); do
+          echo $rules_sets | jq -r ".\"$rules_set\"" > '${globals.rules_dir}'/$rules_set.cfg
+        done
```
laurentfufu, line 40: I don't understand how this can works because after line 31, `rules_sets` while be something like this:
```json
{
  "rewrite-rules": "{\"file_1\": \"# this \\n# is file 1\", \"file_2\":\"# and this\\n# is file 2 and I □ unicode\"}"
}
```

so `echo $rules_sets | jq -r 'keys[]'` will output only `rewrite-rules` and we will have a a `rewrite-rules` file name with the content this content:
```
{"file_1": "# this \n# is file 1", "file_2":"# and this\n# is file 2 and I □ unicode"}
```

??

laurentfufu (changes requested):

packages/jahia/rewrite-rules.yml
laurentfufu: forget, i'm nut, there will be no `rewrite-rules` key anymore, sorry ! :trollface:

packages/jahia/rewrite-rules.yml
```diff
@@ -24,24 +21,36 @@ actions:
...
         fi
 
-        if echo "${settings.md5sum} ${globals.temp_rules_file}" | md5sum --check --status ; then
-          mv -f ${globals.temp_rules_file} ${globals.rules_file}
-        else
-          echo "MD5 sum mismatch. Aborting" >&2
-        fi
+        # Get all rules secret and store as individual files
+        for rules_set in $(echo $rules_sets | jq -r 'keys[]'); do
+          echo $rules_sets | jq -r ".\"$rules_set\"" > '${globals.rules_dir}'/$rules_set.cfg
```
laurentfufu, line 39: backslash escapes will stay literals here

laurentfufu (dismissed):

packages/jahia/rewrite-rules.yml
amitc-jahia: Changes have been made. Backslash escapes are being handled

Checked max line command size with config around 100k lines and it is not an issue.

adoux (dismissed):

adoux:
Approved, waiting for PAAS-1576 to be merged before merging that one

kevinvandel (dismissed):
@amitc-jahia an action is needed in the migration package to update the file structure on existing envs, otherwise when we release we would break the feature for all existing envs with a custom conf (so probably all prod envs).
Don't hesitate to reach out to Cyril if you need insight on that part.

amitc-jahia:
Migration package updated to handle custom rules for existing envs.

adoux (dismissed):

adoux:
@amitc-jahia, sorry... But conflicts :grimacing:

amitc-jahia:
Rebased

kevinvandel:
@amitc-jahia could you please squash the 2 last commits (6419c4a and 07933e9) instead of having one with ugly conflicts and another to remove them ? (you can do that with an interactive rebase or a reset HEAD).
Also it would be nice to move the `updateHAProxyRulesFile` action up in the migration package, I tried to order all actions so that we have :
1. actions that don't require anything special
2. actions that require a restart
3. restart

And I think we need to be careful about the `updateHaproxyConfig` action, which also changes the HAProxy config (and reloads it).

kevinvandel (changes requested):
@amitc-jahia could you please squash the 2 last commits (6419c4a and 07933e9) instead of having one with ugly conflicts and another to remove them ? (you can do that with an interactive rebase or a reset HEAD).
Also it would be nice to move the `updateHAProxyRulesFile` action up in the migration package, I tried to order all actions so that we have :
1. actions that don't require anything special
2. actions that require a restart
3. restart

And I think we need to be careful about the `updateHaproxyConfig` action, which also changes the HAProxy config (and reloads it).

kevinvandel:
> And I think we need to be careful about the `updateHaproxyConfig` action, which also changes the HAProxy config (and reloads it).

@amitc-jahia I just had a closer look, and your action needs to be done before `updateHaproxyConfig`, because the filename matters to this action (so the renaming needs to be done beforehand).

amitc-jahia:
Changes made as requested regarding the commits and action in the migration package.

Also moved the action for [https://jira.jahia.org/browse/PAAS-1690](https://jira.jahia.org/browse/PAAS-1690) in the migration package before the restarts.

Each commit in the PR is a dedicated action.

kevinvandel (approved):

PR #112, opened 2021-08-09
PAAS-1694 IPsec datadog custom check handle multi-connections with sa…

laurentfufu:
…me rightsubnet

JIRA issue: https://jira.jahia.org/browse/PAAS-1694

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v8.yaml
+++ b/packages/jahia/migrations/migrate-to-v8.yaml
@@ -36,7 +36,8 @@ onInstall:
   - enableNativeMemoryTracking                             # PAAS-1705
   - configureDatadogAPM                                    # PAAS-1712
   - disableIoCheck                                         # PAAS-1542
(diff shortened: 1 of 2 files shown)


packages/jahia/migrations/migrate-to-v8.yaml
```diff
@@ -36,6 +36,7 @@ onInstall:
   - enableNativeMemoryTracking                             # PAAS-1705
   - configureDatadogAPM                                    # PAAS-1712
   - disableIoCheck                                         # PAAS-1542
+  - updateDDIPSecConnStatus                                # PAAS-1694
```
kevinvandel, line 39: IF I'm not mistaken, PAAS-1694 also needs to be added to the list of tickets needing `restartDDAgentOnJahiaNodes` (in the comment on the line below)

laurentfufu: you're damn right, fixed

kevinvandel (approved):

adoux (approved):

PR #119, opened 2021-08-20
PAAS-1715: Refresh jCustomer allowed IPs on Jahia browsing scale in/out

kevinvandel:
JIRA issue: https://jira.jahia.org/browse/PAAS-1715

Short description:
I also added the action at jCustomer env restart in case the env was stopped during scaling of a linked Jahia env.

### DIFF
--- a/packages/jcustomer/install.yml
+++ b/packages/jcustomer/install.yml
@@ -15,14 +15,6 @@ description:
 globals:
   jcustomer_env_version: 7
   unomi_version: ${settings.productVersion}
-  stack_connect_ip_adds:
-    - 3.65.225.246
-    - 3.66.45.94
-    - 18.198.249.58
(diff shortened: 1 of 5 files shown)


packages/jcustomer/install.yml
```diff
@@ -31,101 +23,103 @@ mixins:
...
-      ES_VERSION: esVersion,
-      envmode: envmode,
-      ORGANIZATION_NAME: organizationName
-    },
-    "distribution": {
-        "mode": "STRICT",
-        "zones": [
-            {high_iops: false}
+onBeforeInstall:
+  - getWorkatoIps
```
kevinvandel, line 27: This is the only thing I added, all the rest of the diff comes from the needed reindent :grimacing:

Ronan-Gorain (approved):

PR #120, opened 2021-08-25
PAAS-1732: fix infinite loop in installModuleIfNotPresent action

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-1732

Short description:


mixins/jahia.yml
```diff
@@ -561,6 +561,7 @@ actions:
                   'NF==5 && $1>0 && $2!="Active" && $4=="${globals.moduleVersion}" && $NF~/^${globals.moduleName}$/ \
                   {print $1}')"
         if [ -n "$bundle_id" ]; then
+          imax=66
```
kevinvandel, line 564: To be consistent with PAAS-1731, I suggest we set 120 here (to wait 2 minutes tops)

kevinvandel (changes requested):

kevinvandel (approved):

PR #134, opened 2021-09-07
Paas 1736 mixins vault

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1736

Short description:

### DIFF
--- a/mixins/common.yml
+++ b/mixins/common.yml
@@ -320,3 +320,122 @@ actions:
         curl -fLSso /usr/local/bin/papi.py ${globals.repoRootUrl}/assets/common/papi.py || exit 1
         chmod 755 /usr/local/bin/papi.py
       user: root
+
+  ###################
+  # Vault actions   #
+  ###################
+  vaultAppRoleLogin:
+    # Returns:
+    #   ${globals.vaultToken}
+    - set:
+        generateNewToken: true
+    - if ("${globals.vaultToken.print()}" != ""):
+        - cmd [${this}}]: |-
+            curl -s -H "X-Vault-Token: ${globals.vaultToken}" $VAULT_CLUSTER_URL/v1/auth/token/lookup-self \
+              | jq -er .data &>/dev/null
+            if [ $? -ne 0 ]; then
+              echo "Token is not valid, it needs to be renewed"
+            fi
+        - if ("${response.out" == ""):
+            set:
+              generateNewToken: false
+    - if ("${this.generateNewToken}" == "true"):
+        - cmd [${this}]: |-
+            payloadfile=$(mktemp)
+            cat > $payloadfile <<END
+            {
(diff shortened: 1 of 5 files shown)


mixins/common.yml
```diff
@@ -320,3 +320,122 @@ actions:
...
+              "secret_id": "$VAULT_SECRET_ID"
+            }
+            END
+            VAULT_TOKEN=$(curl -s -XPOST $VAULT_CLUSTER_URL/v1/auth/approle/login --data @$payloadfile | jq -er .auth.client_token)
+            if [ $? -ne 0 ]; then
+              echo "[ERROR] Can't generate vault token with appRole login (VAULT_TOKEN=$VAULT_TOKEN)" >&2
+              exit 1
+            fi
+            echo $VAULT_TOKEN
+            rm -f $payloadfile
```
Ronan-Gorain, line 357: This should also be in the previous if block to avoid having another file with credentials in the filesystem.

mixins/common.yml
```diff
@@ -320,3 +320,122 @@ actions:
         curl -fLSso /usr/local/bin/papi.py ${globals.repoRootUrl}/assets/common/papi.py || exit 1
         chmod 755 /usr/local/bin/papi.py
       user: root
+
+  ###################
+  # Vault actions   #
+  ###################
+  vaultAppRoleLogin:
```
Ronan-Gorain, line 327: It takes a node/nodegroup as parameters. It might help to add it to the description.

packages/jahia/rewrite-rules.yml
```diff
@@ -6,12 +6,17 @@ id: jahia-set-rewrite-rules
...
+  - "../../mixins/haproxy.yml"
+
 globals:
   rules_dir: "/etc/haproxy/haproxy.cfg.jahia/customer.configuration.d"
   backup_rules_dir: "/var/lib/haproxy/customer.configuration.d"
   temp_rules_dir: "/var/lib/haproxy/customer.configuration.tmp"
 
 onInstall:
+  - vaultGetRewriteRulesB64
   - getRewritesRules
```
Ronan-Gorain, line 20: Maybe it's time to rename the getRewritesRules action into setRewritesRules as it's not even fetching rules anymore.

Ronan-Gorain (changes requested):

mixins/common.yml
adoux: Good catch, fixed

mixins/common.yml
adoux: We usually don't add any comment when it's a single parameter but I don't mind adding it. Done

packages/jahia/rewrite-rules.yml
adoux: Smart indeed. Done!

Ronan-Gorain (approved):

PR #139, opened 2021-09-09
PAAS-1736: add recursive mode to vaultSecretsList action

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1736

Short description:

### DIFF
--- a/mixins/common.yml
+++ b/mixins/common.yml
@@ -441,16 +441,11 @@ actions:
           echo $resp | jq -cre '.data.keys | join("\n")' > $tmp_secrets_list
           cat $tmp_secrets_list | while read path_end; do
             full_path="$path/$path_end"
-            # If "secret" ends with a slash, means it's a subpath
-            if echo $path_end | grep -qE "/$"; then
-              # If recursive, then we list the secrets under this subpath
-              # else we just print the subpath with the trailing slash
-              if [ "$recursive" = "true" ]; then
-                # We remove the trailing slash from the full path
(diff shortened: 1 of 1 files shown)


mixins/common.yml
```diff
@@ -411,23 +411,57 @@ actions:
...
+              # else we just print the subpath with the trailing slash
+              if [ "$recursive" = "true" ]; then
+                # We remove the trailing slash from the full path
+                list_secrets_to_file $(echo -n $full_path | head -c -1)
+              else
+                echo $full_path >> $final_secrets_list
+              fi
+            else
+              echo $full_path >> $final_secrets_list
+            fi
```
Ronan-Gorain, line 456:
```suggestion
            # If recursive and "secret" ends with a slash, means it's a subpath
            #     We remove the trailing slash from the full path
            # else we just print the subpath with the trailing slash
            if [ "$recursive" = "true" ] && echo $path_end | grep -qE "/$"; then
                list_secrets_to_file $(echo -n $full_path | head -c -1)
            else
              echo $full_path >> $final_secrets_list
            fi
```

Ronan-Gorain: It's just a suggestion (tested). If you disagree, I'll merge.

Ronan-Gorain: I'm also not sure why the comment talks about removing the trailing slash from the full path as I don't see this operation on lines arround

adoux: it's the line right after (`head -c -1`) :-)

Ronan-Gorain (approved):

PR #142, opened 2021-09-14
PAAS-1760: install jq and python everywhere at env creation

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1760

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v9.yaml
+++ b/packages/jahia/migrations/migrate-to-v9.yaml
@@ -32,6 +32,7 @@ onInstall:
   - datadogCustomCheckForJahiaNodes                # PAAS-1693
   - installPapiScript: bl, proc, cp, sqldb         # PAAS-1686
   - updateProxySQLConfigForSingleDBNodes           # PAAS-1647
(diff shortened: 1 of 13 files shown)


packages/jahia/migrations/migrate-to-v9.yaml
```diff
@@ -64,3 +65,8 @@ actions:
   updateProxySQLConfigForSingleDBNodes:
     if (nodes.sqldb.length == 1):
       configureProxysqlForSingleDBNode
+
+  installRequiredPackagesEverywhere:
+    - installRequiredPackages:
+        target: "cp, es"
```
Ronan-Gorain, line 71: This is more "cp, proc, bl, sqldb" here

Ronan-Gorain (changes requested):

packages/jahia/migrations/migrate-to-v9.yaml
adoux: Very good call :man_facepalming: 
Fixed!

Ronan-Gorain (approved):

PR #147, opened 2021-09-20
Paas 1723

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-1723

Short description:

### DIFF
--- a/mixins/mariadb.yml
+++ b/mixins/mariadb.yml
@@ -164,6 +164,7 @@ actions:
         - cmd [${this.target}]: |-
             echo "SERVICE=''" > /var/lib/jelastic/overrides/envinfo.lib
             echo "/var/lib/jelastic/overrides/envinfo.lib" >> /etc/jelastic/redeploy.conf
+          user: root
 
   ##################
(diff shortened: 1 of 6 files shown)


laurentfufu:
should be rebased :)

mixins/mariadb.yml
```diff
@@ -308,38 +326,60 @@ actions:
...
+          - setGlobals:
+              origMasterId: ${@i.id}
+
+    - script: |-
+        masterIndex = "${globals.galeraMasterIndex}".slice(-1);
+        newMasterIndex = ((masterIndex + 1) % 3) + 1
+        return {'result':0, 'newMasterIndex': newMasterIndex}
+
+    - proxysqlSwitchMaster:
+        - target: ${response.newMasterIndex}
```
adoux, line 354: Should not be a list element, right?

Ronan-Gorain: No it should be a number, but isn't it the case here ?

adoux (approved):

PR #152, opened 2021-09-22
PAAS-1764: add vault to jcustomer environments

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-1764

Short description:

### DIFF
--- a/packages/jcustomer/migrations/migrate-to-v8.yaml
+++ b/packages/jcustomer/migrations/migrate-to-v8.yaml
@@ -28,6 +28,7 @@ onInstall:
 
   - installPapiScript: cp, es               # PAAS-1686
   - refreshUnomiAllowedIPs: cp              # PAAS-1735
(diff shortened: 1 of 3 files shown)


packages/jcustomer/migrations/migrate-to-v8.yaml
```diff
@@ -68,3 +69,91 @@ actions:
...
+        splitted_output = output.split(",")
+        return {
+          'result': 0,
+          "role_id":splitted_output[0],
+          'secret_id':splitted_output[1]
+        }
+    - setGlobals:
+        role_id: ${response.role_id}
+        secret_id: ${response.secret_id}
+
```
kevinvandel, line 124: It doesn't seem necessary to define those variables as global, so I think you can use `set` instead of `setGlobal` (and then `${this.myVar}` instead of `${globals.myVar}`)

adoux (approved):

PR #174, opened 2021-10-20
Paas 1810 fix rewrite rules

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1810

Short description:
I added the ability to gzip the secret data even though we won't use it for fetching rewrite rules, could still be useful


kevinvandel (changes requested):
Not sure about releasing a "dirty" hotfix, since there is a workaround: splitting the conf in multiple files. We need to discuss that before merging anything, that's why I'm blocking the PR.

packages/jahia/rewrite-rules.yml
```diff
@@ -30,9 +29,17 @@ actions:
           mv -f ${globals.rules_dir}/* ${globals.backup_rules_dir}
         fi
 
+        # Connect to Vault and check if rules exist
```
kevinvandel, line 32: @adoux it would be nice to add a massive warning here to explain that we do it like that because using an action (that sets a global) doesn't work in the case the config is massive

packages/jahia/rewrite-rules.yml
```diff
@@ -16,7 +16,6 @@ globals:
   temp_rules_dir: "/var/lib/haproxy/customer.configuration.tmp"
 
 onInstall:
-  - vaultGetRewriteRulesB64
```
kevinvandel, line 19: We should completely delete this action, it's not used anymore, and worse: it MUST NOT be used

mixins/common.yml
```diff
@@ -376,9 +376,13 @@ actions:
     # Parameters:
     #   - target: target node group or id
     #   - secretPath: path (and name) of the secret to fetch
+    #   - gzip: if the secret data should be gzipped or not
     # Returns:
     #   ${globals.vaultSecretData}: A JSON string with all key and data for the specified secret, base64 encoded
     - vaultAppRoleLogin: ${this.target}
+    - if ("${this.gzip.print()}" != "true"):
+        set:
+          gzip: false
```
kevinvandel, line 385: it was a good try, but since the gzip thing doesn't work at all, we should not keep it

PR #205, opened 2021-11-26
PAAS-1739: Improve Jahia/jCustomer link management when Jahia is deleted

amitc-jahia:
JIRA issue: https://jira.jahia.org/browse/PAAS-1739

Short description:

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -338,17 +338,12 @@ actions:
   deleteEnvLink:
     - script: |-
         const envName = "${env.envName}";
-        jCustomerEnv = jelastic.env.control.GetNodeGroups(envName, session).object.filter(function (object) {
-                                return object.name == "cp";
-                              }).pop().envLink;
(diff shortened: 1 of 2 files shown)


adoux (approved):

mixins/jahia.yml
```diff
@@ -338,17 +338,12 @@ actions:
   deleteEnvLink:
     - script: |-
         const envName = "${env.envName}";
-        jCustomerEnv = jelastic.env.control.GetNodeGroups(envName, session).object.filter(function (object) {
-                                return object.name == "cp";
-                              }).pop().envLink;
-        if (jCustomerEnv == null) {
-          return {"result": 0, "out": "No jCustomer env linked"};
-        }
+        const jCustomerEnv = "${globals.unomi_env_name}";
```
kevinvandel, line 341: This should be forbidden, we shouldn't use globals that are defined out of the scope of the action. We need to pass it as parameter of the action instead.

PR #210, opened 2021-11-29
PAAS-1870: fix hn upgrade script

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1870
Short description:


assets/common/hn_upgrade.py
```diff
@@ -55,7 +56,7 @@ def start_hn_upgrade(self):
         valid_envnames = get_jahia_cloud_envnames_from_papi(self.papi_hostname, self.papi_token)
         running_containers = self.get_running_containers_id(jelastic_session, hn_infos["id"], valid_envnames)
 
-        if not self.recover_state:
+        if self.recover_state and not self.no_update_state_file:
```
Ronan-Gorain, line 59: I'm not sure to understand the necessity of this new parameter which seems to be a duplicate of "recover_state",  which is already used to know the user wants to restore the list or not (and if it restores it, no need to save it first so this step was skipped).

Ronan-Gorain (commented):
Good catch :) 
We just need to clarify the no_update_state_file parameter before I merge

assets/common/hn_upgrade.py
adoux: Oooopps my bad, was totally mistaken, didn't read the algo properly... 🤦 
Updated the PR

Ronan-Gorain (approved):

PR #212, opened 2021-11-30
PAAS-1739: Fix actions to use globals and add migration package to up…

amitc-jahia:
…date-events

JIRA issue: https://jira.jahia.org/browse/PAAS-1739

Short description:

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -336,9 +336,13 @@ actions:
         checkName: strongswan_connections_status
 
   deleteEnvLink:
+  # Parameters:
+  #   - jCustomerEnv: Jcustomer env name in the env link of Jahia
+  # Returns:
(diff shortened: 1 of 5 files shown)


mixins/jahia.yml
```diff
@@ -338,7 +338,7 @@ actions:
   deleteEnvLink:
```
kevinvandel, line 338: Could you please add a doc-comment specifying the parameter (`jCustomerEnv`) and the response (`${response.link_removed}`), like we do for other actions ?

kevinvandel (changes requested):

mixins/jahia.yml
amitc-jahia: Done

kevinvandel (approved):

laurentfufu (approved):

PR #213, opened 2021-12-01
PAAS-1740: improve jahia/jcustomer envlink at jcustomer deletion

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-1740

Short description:


kevinvandel (dismissed):
We need to add an action in the Jahia migration package to remove all empty `envLink`

Ronan-Gorain:
Done

adoux (approved):

PR #215, opened 2021-12-01
PAAS-1846 enable jahia's java process metrics on datadog-agent

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1846
Short description:


kevinvandel (changes requested):
migration package v10 is related to the previous release, so we need to start a new one

laurentfufu:
> migration package v10 is related to the previous release, so we need to start a new one

stupid me, sorry! :grin:

laurentfufu:
we will apply this with image redeploy after all, so no need to use the migration script anymore

PR #216, opened 2021-12-03
PAAS-1865: Mute/unmute synthetics at environment stop/start/upgrade/r…

amitc-jahia:
…estore

JIRA issue: https://jira.jahia.org/browse/PAAS-1865

Short description:


mixins/common.yml
```diff
@@ -688,3 +688,69 @@ actions:
...
+        check_monitor_status=$(curl -s $DD_MONITOR_SEARCH \
+                                -H 'Content-Type: application/json' \
+                                -H "DD-API-KEY: $DATADOGAPIKEY" \
+                                -H "DD-APPLICATION-KEY: $DD_APP_KEY" \
+                                | jq -er '.monitors[] | select(.id | contains('$monitor_id')) | .muted_until_ts')
+        if [ $check_monitor_status != null ]; then
+          echo "[INFO] Synthetics is already muted. Nothing to do."
+          exit 0
+        fi
+        curl -s "https://api.datadoghq.com/api/v1/monitor/$monitor_id/mute" \
```
Ronan-Gorain, line 719: You should add an expiration to the timeout. Even if it's 24h or something else big, it will avoid to forget that the monitor is muted  in case the package can't reach the end of the execution, but the env is working.

Ronan-Gorain (changes requested):

Ronan-Gorain (approved):

PR #217, opened 2021-12-07
COPS-478 fix dd's haproxy integration conf which miss service and sta…

laurentfufu:
…tus checks

JIRA issue: https://jira.jahia.org/browse/COPS-478

Short description:


kevinvandel:
@laurentfufu please move the ticket to PAAS project, set the right "component", and then open a new PR with the correct ticket ID.
Also I guess we need to add an action in the migration package to fix existing environments.

laurentfufu:
done here: Jahia/jelastic-packages#219

PR #218, opened 2021-12-08
PAAS-1891: Fix action to add proc node for onAfterRedeployContainer e…

amitc-jahia:
…vent

JIRA issue: https://jira.jahia.org/browse/PAAS-1891

Short description:

### DIFF
--- a/packages/jahia/update-events.yml
+++ b/packages/jahia/update-events.yml
@@ -115,14 +115,12 @@ onAfterRedeployContainer[proc]:
         user: root
       - if ('${response.out}' == ''):
           - startupJahiaHealthCheck: ${event.params.nodeGroup}
-          - addProcNodeToHaproxy:
-              reload: true
   - elif ('${event.params.nodeId:}'):
       # Handle RedeployContainerById
       - onAfterRedeployJahiaContainer: ${event.params.nodeId}
       - startupJahiaHealthCheck: ${event.params.nodeId}
-      - addProcNodeToHaproxy:
-          reload: true
+  - addProcNodeToHaproxy:
+      reload: true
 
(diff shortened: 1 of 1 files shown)


packages/jahia/update-events.yml
```diff
@@ -115,8 +115,8 @@ onAfterRedeployContainer[proc]:
         user: root
       - if ('${response.out}' == ''):
           - startupJahiaHealthCheck: ${event.params.nodeGroup}
-          - addProcNodeToHaproxy:
-              reload: true
+      - addProcNodeToHaproxy:
+          reload: true
```
laurentfufu, line 119: what about this instead:
```yaml
onAfterRedeployContainer[proc]:
  - if ('${event.params.nodeGroup:}'):
      # Handle RedeployContainersByGroup
      - onAfterRedeployJahiaContainer: ${event.params.nodeGroup}
      - cmd [proc]: |-
          echo $JAHIA_UPGRADE
        user: root
      - if ('${response.out}' == ''):
          - startupJahiaHealthCheck: ${event.params.nodeGroup}
  - elif ('${event.params.nodeId:}'):
      # Handle RedeployContainerById
      - onAfterRedeployJahiaContainer: ${event.params.nodeId}
      - startupJahiaHealthCheck: ${event.params.nodeId}
  - addProcNodeToHaproxy:
      reload: true
```

laurentfufu (changes requested):

packages/jahia/update-events.yml
amitc-jahia: Much better. Changed. Thanks

adoux (approved):

PR #222, opened 2021-12-09
PAAS-1865: Add check to unmute after monitor status is OK

amitc-jahia:
JIRA issue: https://jira.jahia.org/browse/PAAS-1865

Short description:

### DIFF
--- a/mixins/common.yml
+++ b/mixins/common.yml
@@ -704,18 +704,18 @@ actions:
                       -H 'Content-Type: application/json' \
                       -H "DD-API-KEY: $DATADOGAPIKEY" \
                       -H "DD-APPLICATION-KEY: $DD_APP_KEY" \
-                      | jq -er '.tests[] | select(.config.request.url | contains("${env.shortdomain}")) | .monitor_id')
+                      | jq -er '.tests[] | select(.config.request.url | contains("https://${env.shortdomain}.")) | .monitor_id')
         if [[ $? -ne 0 ]]; then
           echo "[INFO] No synthetics defined for ${env.envName}"
           exit 0
         fi
         DD_MONITOR_SEARCH=https://api.datadoghq.com/api/v1/monitor/search?query=synthetics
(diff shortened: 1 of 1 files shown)


mixins/common.yml
```diff
@@ -738,21 +738,32 @@ actions:
...
+        monitor_mute_status=$(curl -s $DD_MONITOR_SEARCH \
+                             -H 'Content-Type: application/json' \
+                             -H "DD-API-KEY: $DATADOGAPIKEY" \
+                             -H "DD-APPLICATION-KEY: $DD_APP_KEY" \
+                             | jq -er '.monitors[] | select(.id | contains('$monitor_id')) | .muted_until_ts')
+        if [ $monitor_mute_status == null ]; then
           echo "[INFO] Synthetics is already unmuted. Nothing to do."
           exit 0
         fi
+        for (( i=1; i<=180; i++ )); do
```
kevinvandel, line 756: Could you please add a comment to explain what we're doing here ? The whole action is getting quite complex, so having small comments summing up each step would we really cool :)

kevinvandel (changes requested):

mixins/common.yml
amitc-jahia: Done

amitc-jahia: The 180 is the number of iteration and not actually seconds. Along with the 1s sleep in the loop, the total time for the action to complete takes 5mins. If I have 300 iterations then it takes 8 mins, which is way too longer.

kevinvandel: Damn, so that means the API call takes almost 700ms ?! Sounds like a lot :grimacing: 
Anyway, thanks for the explanation. It could be worth being more accurate in the comment then, like `Once max iterations (~5mins for 180 iterations) is reached`.

kevinvandel (dismissed):

mixins/common.yml
amitc-jahia: Changed

adoux (approved):

PR #227, opened 2021-12-14
Paas 1740

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-1740

Short description:

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -889,24 +889,13 @@ actions:
             type: error
             message: "jCustomer password is wrong."
 
-  removeJexperienceConfigIfNoEnvLink:
-    - getEnvLinkedJcustomer:
-        envName: ${env.envName}
-    - if ( ! ${globals.unomi_env_linked}):
-      - cmd[proc, cp]: |-
-          JCUSTOMER_CONFIG_FILE="/data/digital-factory-data/karaf/etc/org.jahia.modules.jexperience.settings-global.cfg"
-          if [ -f $JCUSTOMER_CONFIG_FILE ]; then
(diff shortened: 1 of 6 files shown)


mixins/jahia.yml
```diff
@@ -889,6 +889,17 @@ actions:
...
 
+  removeJexperienceConfigIfNoEnvLink:
+    - getEnvLinkedJcustomer:
+        envName: ${env.envName}
+    - if ( ! ${globals.unomi_env_linked}):
+      - cmd[proc, cp]: |-
+          JCUSTOMER_CONFIG_FILE="/data/digital-factory-data/karaf/etc/org.jahia.modules.jexperience.settings-global.cfg"
+          if [ -f $JCUSTOMER_CONFIG_FILE ]; then
+            sed -i 's/\(jexperience.jCustomer.*=\).*/\1/g' $JCUSTOMER_CONFIG_FILE
+            rm -f /data/digital-factory-data/modules/jexperience-*.jar
```
kevinvandel, line 900: If we could avoid code duplication with the `removeJexperienceConfig` action, it would be really cool.

Ronan-Gorain: It's possible but it will require to refactor many things and make things less clear in my opinion, so as you prefer.

kevinvandel: Yes some refactoring will be needed (and I just saw that we should refactor the `unlink-jahia-jcustomer` package too), but not sure it would make things less clear (but you can add comments if needed to make things clearer).
Also I think we should rename the action `removeJexperienceConfig`, because it does not only remove the config (and actually it does not even remove it).

adoux (approved):

PR #233, opened 2021-12-20
PAAS-1919: add package to update datadog agent on all nodes

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-1919

Short description:

### DIFF
--- a/packages/common/update-datadog-agent.yml
+++ b/packages/common/update-datadog-agent.yml
@@ -1,0 +1,15 @@
+---
+type: update
+version: 1.5
+name: Update datadog agent
+logo: ../../assets/common/jahia-logo-70x70.png
+id: update-datadog-agent
+
+
+onInstall:
+  cmd[*]: |-
+    source /.jelenv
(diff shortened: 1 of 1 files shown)


packages/common/update-datadog-agent.yml
```diff
@@ -0,0 +1,15 @@
+---
+type: update
+version: 1.5
+name: Jahia - update datadog agent
+logo: ../../assets/common/jahia-logo-70x70.png
+id: jahia-update-datadog-agent
```
kevinvandel, line 6: I think we can drop the "Jahia" prefix, to me it makes sense only for packages specific to Jahia envs (but it doesn't seem to be true of all "common" packages...)

Ronan-Gorain: > but it doesn't seem to be true of all "common" packages...
Yes that's why I kept it, I thought it was about the jahia company :) 

Anyway, I updated the package

PR #236, opened 2021-12-21
PAAS-1737 fix some conditionals, always set variable even if false

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1737

Short description:

### DIFF
--- a/packages/common/restore.yml
+++ b/packages/common/restore.yml
@@ -114,7 +114,7 @@ actions:
                 printf "{\"version\": \"%s\", \"ACTIVE\": %s }",version,ACTIVE \
               } \
               else{ \
-              printf "{\"ACTIVE\": false}" \
+                printf "{\"ACTIVE\": false}" \
               } \
             }' \
           /data/digital-factory-data/bundles-deployed/*/bundle.info
@@ -133,7 +133,7 @@ actions:
         step: "before restore"
     - setGlobals:  # must be set for following conditionnal test, even if false
         jexperience_active: ${response.ACTIVE}
-    - if (globals.jexperience_active):
+    - if (globals.jexperience_active == "true"):
         - setGlobals:
             jexperience_active_version: ${response.version}
         # we keep the jexperience conf in a barely safe spot
@@ -193,9 +193,11 @@ actions:
     - elif (!${globals.jexperience_active} && ${this.restored_linked}):  # we don't have to keep the restored jexperience
         - setGlobals:
             remove_restored_jexperience: true
+            reinstall_jexperience: false
             same_jexperience_conf: false
     - elif (${globals.jexperience_active} && !${this.restored_linked}):  # we have to reinstall jexperience module
         - setGlobals:
+            remove_restored_jexperience: false
             reinstall_jexperience: true
             same_jexperience_conf: false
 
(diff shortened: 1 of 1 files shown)


packages/common/restore.yml
```diff
@@ -311,7 +313,8 @@ actions:
         - disableKarafLogin: cp, proc
 
     # we remove the jexperience backup, if any left
-    - if (globals.jexperience_active):
+    # be awre that this global is a string
```
kevinvandel, line 316: typo :stuck_out_tongue:

laurentfufu: fixed

kevinvandel (changes requested):
I think there are other test conditions to fix:
https://github.com/Jahia/jelastic-packages/blob/b739217ee91313d70c5b89efa7ad6aec6a57b2a4/packages/common/restore.yml#L279
https://github.com/Jahia/jelastic-packages/blob/b739217ee91313d70c5b89efa7ad6aec6a57b2a4/packages/common/restore.yml#L297
https://github.com/Jahia/jelastic-packages/blob/b739217ee91313d70c5b89efa7ad6aec6a57b2a4/packages/common/restore.yml#L306

EDIT: I was wrong, in those cases the variable really contains a boolean and not a string :exploding_head:

packages/common/restore.yml
```diff
@@ -193,11 +193,18 @@ actions:
...
     - elif (${globals.jexperience_active} && !${this.restored_linked}):  # we have to reinstall jexperience module
         - setGlobals:
+            remove_restored_jexperience: false
             reinstall_jexperience: true
             same_jexperience_conf: false
+    - elif (!${globals.jexperience_active} && !${this.restored_linked}):  # nothing to do
+        - setGlobals:
+            remove_restored_jexperience: false
+            reinstall_jexperience: false
+            same_jexperience_conf: false
```
kevinvandel, line 207: I'm not 100% sure, but I think those 3 globals can be defined locally instead of globally (`set` instead of `setGlobals`).

kevinvandel (changes requested):

kevinvandel:
>

laurentfufu:
> I think there are other test conditions to fix:
> 
> https://github.com/Jahia/jelastic-packages/blob/b739217ee91313d70c5b89efa7ad6aec6a57b2a4/packages/common/restore.yml#L279
> 
> https://github.com/Jahia/jelastic-packages/blob/b739217ee91313d70c5b89efa7ad6aec6a57b2a4/packages/common/restore.yml#L297
> 
> https://github.com/Jahia/jelastic-packages/blob/b739217ee91313d70c5b89efa7ad6aec6a57b2a4/packages/common/restore.yml#L306
> 
> EDIT: I was wrong, in those cases the variable really contains a boolean and not a string exploding_head

now using the famous `jelastic don't care about bolean or string type` form :trollface:

packages/common/restore.yml
laurentfufu: globals are required here because there are used in conditional statements later in the manifest

kevinvandel (dismissed):

kevinvandel (approved):

PR #238, opened 2021-12-22
PAAS-1887: Fix jExperience module upgrade only if already present

amitc-jahia:
JIRA issue: https://jira.jahia.org/browse/PAAS-1887

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v11.yaml
+++ b/packages/jahia/migrations/migrate-to-v11.yaml
@@ -82,9 +82,16 @@ actions:
         b: 8.0.0.0
         res: isLowerThan8000
     - if( ${globals.isLowerThan8000} ):
-        - upgradeJahiaModule:
-            module: "jExperience"
-            version: 1.11.7
+        - enableKarafLogin: proc
+        - cmd[${nodes.proc.first.id}]: |-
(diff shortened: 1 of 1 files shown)


packages/jahia/migrations/migrate-to-v11.yaml
```diff
@@ -82,9 +82,16 @@ actions:
...
         res: isLowerThan8000
     - if( ${globals.isLowerThan8000} ):
-        - upgradeJahiaModule:
-            module: "jExperience"
-            version: 1.11.7
+        - enableKarafLogin: proc
+        - cmd[${nodes.proc.first.id}]: |-
+            ${globals.karafConsole} bundle:list --no-format 2> /dev/null | awk '$NF~/^jExperience$/ {print $1}'
+        - setGlobals:
+            moduleId: ${response.out}
```
kevinvandel, line 89: You don't need a global here, so you can use `set` instead of `setGlobals` :grin: 
(I think it's better to get used to using globals only when needed - which should really not happen very often)

amitc-jahia: Fixed

kevinvandel (approved):

Ronan-Gorain (approved):

PR #242, opened 2022-01-13
PAAS-1937 upgrade Jahia Macros module (security issue)

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1937

Short description:

### DIFF
--- a/packages/one-shot/paas-1937-jahia-macros-patch-security-issue.yml
+++ b/packages/one-shot/paas-1937-jahia-macros-patch-security-issue.yml
@@ -1,0 +1,74 @@
+---
+type: update
+version: 1.5.2
+name: PAAS-1937 - Upgrade Jahia Macros module to patch security issue
+id: paas-1937-upgrade-jahia-macros-module-to-patch-security-issue
+
+mixins:
+  - "../../mixins/common.yml"
+  - "../../mixins/jahia.yml"
+
+onInstall:
+  - getJahiaVersion
+  - isVersionStrictlyLower:
+      a: ${globals.jahiaVersion}
+      b: 7.3.10.0
+      res: upgradeJahiaMacro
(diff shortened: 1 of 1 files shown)


packages/one-shot/paas-1937-jahia-macros-patch-security-issue.yml
```diff
@@ -0,0 +1,70 @@
...
+        jm_v=${globals.jm_v}
+        url="https://devtools.jahia.com/nexus/content/repositories/jahia-releases/org/jahia/modules/macros"
+        oldJMLocalID=$(${globals.karafConsole} bundle:list | awk ${this.awkFS} \
+                     '$NF=="Jahia Macros" {print $1}')
+        oldJMClusterID=$(${globals.karafConsole} cluster:bundle-list default | awk ${this.awkFS} \
+                     '$NF=="Jahia Macros" {print $1}')
+        # install and start new Jahia Macros
+        ${globals.karafConsole} cluster:bundle-install -s default $url/$jm_v/macros-$jm_v.jar
+        # uninstall local old Jahia Macros
+        ${globals.karafConsole} bundle:uninstall $oldJMLocalID
```
kevinvandel, line 68: This step shouldn't be required. If it is, that means there may be something fishy happening here... :thinking:

laurentfufu: I switch to `cluster` commands only but still can have weird status like both `cluster:bundle` active, or same ID.
Not automatically use the `-s` switch (_Start the bundle after installation_) the new module on `cluster:bundle-install`, then start it, stop and uninstall the old looks to works every time on fresh environment.

kevinvandel (approved):

Ronan-Gorain (approved):

PR #243, opened 2022-01-14
PAAS-1875 update Distributed Sessions module

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1875

Short description:


kevinvandel (changes requested):
@laurentfufu could you please move this to a dedicated one-shot package ? In the end this will not be part of a release, but will be handled in a dedicated maintenance operation instead.

laurentfufu:
done!

packages/one-shot/paas-1875-upgrade-distributed-sessions.yaml
```diff
@@ -0,0 +1,152 @@
...
+        - else:
+            - upgradeJahiaModule:
+                module: "Distributed Sessions"
+                version: ${globals.distSessVersion}
+            - cmd[proc, cp]: |-
+                service tomcat stop
+              user: root
+            - cmd[proc, cp]: |-
+                service tomcat start
+              user: root
```
kevinvandel, line 152: It's a bit brutal. To lessen the impact, we should remove nodes from HAProxy first, and add them back afterwards (otherwise this will trigger a lot of alerts).
Also I think we should add a "chechHealth" step at the env, to wait for the env to be up before considering the action done.

kevinvandel (changes requested):

packages/one-shot/paas-1875-upgrade-distributed-sessions.yaml
laurentfufu: The manifest now acts in the same way without taking into account the operatingMode.

kevinvandel (approved):

PR #247, opened 2022-01-18
PAAS-1875 ensure new DS is started at cluster level

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1875
Short description:

### DIFF
--- a/packages/one-shot/paas-1875-upgrade-distributed-sessions.yaml
+++ b/packages/one-shot/paas-1875-upgrade-distributed-sessions.yaml
@@ -134,6 +134,7 @@ actions:
             ${globals.karafConsole} cluster:bundle-uninstall default ds-migration
             # also uninstall old distributed session at cluster level
             old_version=$(${globals.karafConsole} cluster:bundle-list -s default | awk ${globals.awkFS} \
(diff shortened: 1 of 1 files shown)


kevinvandel (changes requested):
Looks good, but I think we need to add something to the package, since this operation is very critical (and if anything goes wrong we need to know right away): I think we should add a action at the end of the package to check that only the latest version of the module is installed, and that it is running. Ideally, this action should output some error that would be visible in the `errOut` of Jelastic (so that we can see it in the Ansible logs), and also I thing this action should be run systematically, even in the case the package doesn't update the module because it considers it is already up-to-date.
Maybe a little `sleep` is needed before checking the module status, I don't know if it takes some time for the module to be started.

laurentfufu:
a new action has been added (and still played) to validate that there is only one version of Distributed module installed and that it is the last one and in the Active state.

kevinvandel (changes requested):
Looks good, but I think we should go even further in `checkStat` action and check the module status locally on all nodes too.

laurentfufu:
now checking at cluster and locale levels on every tomcat

kevinvandel (approved):

PR #257, opened 2022-02-14
PAAS-1954: update jahia_cfg_mvnPath during onAfterRedeploy

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1954

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v12.yaml
+++ b/packages/jahia/migrations/migrate-to-v12.yaml
@@ -30,8 +30,9 @@ onInstall:
   - setGlobalRepoRootUrl
   ### End Pre-checks
 
-  - updateTomcatLogsStartPattern  # PAAS-1930
-  - removeOldMailLib  # PAAS-1945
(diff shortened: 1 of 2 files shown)


packages/jahia/migrations/migrate-to-v12.yaml
```diff
@@ -102,8 +95,10 @@ actions:
           echo "[INFO] Lib has already been removed. Nothing to do."
           exit 0
         fi
-        service tomcat stop
         rm -f $lib_path $lib_path_data
-        service tomcat start
```
kevinvandel, line 107: This will not do, because Jahia must be stopped when the lib is removed.

kevinvandel (changes requested):

packages/jahia/migrations/migrate-to-v12.yaml
adoux: Fixed

kevinvandel (approved):

amitc-jahia (approved):

PR #272, opened 2022-03-08
Paas 1959

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-1959

Short description:

### DIFF
--- a/mixins/elasticsearch.yml
+++ b/mixins/elasticsearch.yml
@@ -5,12 +5,12 @@
 actions:
   setupES:
     - getVaultData
-    - runCircleCIPipeline:
+    - runESCircleCIPipeline:
         workflow: terraform_apply
     - createESAccountAndAliases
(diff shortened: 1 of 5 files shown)


mixins/elasticsearch.yml
```diff
@@ -250,7 +250,142 @@ actions:
...
+              workflow_status = result["items"][0]["status"];
+          }
+          timeout -= sleep_interval
+        }
+        if (workflow_status != "success")
+          return {"result": 1, errOut: "CircleCi Pipleline failed"};
+
+        return {"result": 0};
+
+  updateReplica:
```
adoux, line 362: Should we really keep this action and the next one?

Ronan-Gorain: I guess I added them when resolving a conflict... Fixed

adoux (approved):

PR #276, opened 2022-03-11
PAAS-1909: reload secure settings after updating keystore

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1909

Short description:

### DIFF
--- a/mixins/elasticsearch.yml
+++ b/mixins/elasticsearch.yml
@@ -482,6 +482,18 @@ actions:
     - cmd[${nodes.cp.first.id}]: |-
         output_file=$(mktemp)
         ec_admin_credentials=${globals.ecAdminCredentials}
+        # First we need to reload the secure settings to take into account the ney keys in the keystore
+        return_code=$(curl -sS -o $output_file -w '%{http_code}' \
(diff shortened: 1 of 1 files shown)


mixins/elasticsearch.yml
```diff
@@ -482,6 +482,18 @@ actions:
...
+          -H 'Content-Type: application/json' \
+          -u $ec_admin_credentials \
+          -XPOST "https://$UNOMI_ELASTICSEARCH_ADDRESSES/_nodes/reload_secure_settings")
+        if [ $return_code -ne 200 ]; then
+          cat $output_file | tee -a ${this.logsPath}
+          rm -f $output_file
+          exit 1
+        fi
+        sleep 5
+        > $output_file
```
Ronan-Gorain, line 496: I'm not a bash expert so I may be wrong, but won't this line wait for input endlessly ?

laurentfufu: it's ok in bash but I don't see the point in emptying the file anyway

laurentfufu (changes requested):

laurentfufu (approved):

PR #279, opened 2022-03-17
PAAS-1980: delete ec_deployment if there is no jcustomer env linked

Ronan-Gorain:
remaining. Otherwise, remove env indices

JIRA issue: https://jira.jahia.org/browse/PAAS-1980

Short description:

### DIFF
--- a/mixins/elasticsearch.yml
+++ b/mixins/elasticsearch.yml
@@ -65,8 +65,8 @@ actions:
         fi
 
         pipelineId=$(echo $response | jq -r .id)
-        timeout=600 # 10 minutes
-        sleep_interval=10
+        timeout=1200 # 20 minutes
+        sleep_interval=20
         workflow_status="running"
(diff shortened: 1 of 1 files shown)


mixins/elasticsearch.yml
```diff
@@ -291,7 +293,28 @@ actions:
         ec_deployment = JSON.parse(EntityUtils.toString(resp.getEntity()));
 
         if (ec_deployment["mutualized"]) {
-          return {"result": 0, "out": "Mutualized EC deployment, nothing to do"}
+          if (ec_deployment["paas_environments"].length > 1) {
```
kevinvandel, line 296: Both conditions could be inside the same `if` statement (instead of having 2 blocks)

kevinvandel (changes requested):

mixins/elasticsearch.yml
Ronan-Gorain: Fixed

PR #284, opened 2022-03-23
Paas 1977 jcustomer duplicate events

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1977

Short description:

### DIFF
--- a/packages/jcustomer/migrations/migrate-to-v11.yaml
+++ b/packages/jcustomer/migrations/migrate-to-v11.yaml
@@ -23,10 +23,10 @@ onInstall:
   - setGlobalRepoRootUrl
   ### End Pre-checks
 
+  - fixjCustomerEvents           # PAAS-1977  
   - setJelEnvRoleInNodeGroupData # PAAS-1959
   - configureShardsAndReplicas   # PAAS-1992
   - migrateJcustomer2EC          # PAAS-1970
-  - fixjCustomerEvents           # PAAS-1977  
 
   ### Post-actions
   - setEnvVersion: ${globals.version}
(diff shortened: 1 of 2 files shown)


packages/jcustomer/migrations/migrate-to-v11.yaml
```diff
@@ -26,6 +26,7 @@ onInstall:
   - setJelEnvRoleInNodeGroupData # PAAS-1959
   - configureShardsAndReplicas   # PAAS-1992
   - migrateJcustomer2EC          # PAAS-1970
+  - fixjCustomerEvents           # PAAS-1977
```
kevinvandel, line 29: Question: shouldn't we perform this _before_ the "eventsUpdate" action ?

packages/jcustomer/migrations/migrate-to-v11.yaml
```diff
@@ -356,6 +357,57 @@ actions:
           type: error
           message: "envmode value is wrong on ${env.envName}"
 
+  fixjCustomerEvents:
+    - setGlobals:
+        jel_dashboard_url: ${platformUrl}
```
Ronan-Gorain, line 362: This variable seems to be unused

adoux: Good call! Deleted

packages/jcustomer/migrations/migrate-to-v11.yaml
adoux: Nope, events are linked to an installedApp, so when we update the events with the new JPS id, it duplicates events to and associate the new ones with the new installedApp.
When we delete the previous one, it only deletes events that are associated to the previous installedApp

Ronan-Gorain (approved):

PR #291, opened 2022-03-29
PAAS-1989: Fix the check condition

amitc-jahia:
JIRA issue: https://jira.jahia.org/browse/PAAS-1989

Short description:

### DIFF
--- a/mixins/elasticsearch.yml
+++ b/mixins/elasticsearch.yml
@@ -66,7 +66,8 @@ actions:
           exit 1
         fi
 
-        if ! [ $ec_deployment_mutualized ] && [ "$ec_deployment_es_endpoint" != null ]; then
+        # Skip the timeout while updating a mutualized EC deployment but wait till end if it's a new EC deployment
+        if [ "$ec_deployment_mutualized" = "false" ] || [ "$ec_deployment_es_endpoint" = null ]; then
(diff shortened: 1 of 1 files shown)


mixins/elasticsearch.yml
```diff
@@ -66,7 +66,7 @@ actions:
           exit 1
         fi
 
-        if ! [ $ec_deployment_mutualized ] && [ "$ec_deployment_es_endpoint" != null ]; then
+        if ! [ "$ec_deployment_mutualized" = "true" ] && [ "$ec_deployment_es_endpoint" != null ]; then
```
kevinvandel, line 69: While we're at it, I think it could be nice to add a simple comment here to explain what we're doing (wait for CCI worflow's end in some cases but not others)

kevinvandel (dismissed):

mixins/elasticsearch.yml
Ronan-Gorain: This may be more clear to read (no need to negate the conditions) :
```suggestion
        if [ "$ec_deployment_mutualized" = "false" ] || [ "$ec_deployment_es_endpoint" = "null" ]; then
```
I precise that this is not tested :)

amitc-jahia: Done

adoux (approved):

PR #312, opened 2022-04-11
PAAS-2019: don't wait 20m if no workflow created

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2019

Short description:

### DIFF
--- a/mixins/elasticsearch.yml
+++ b/mixins/elasticsearch.yml
@@ -78,6 +78,7 @@ actions:
           echo "Pipeline trigger failed: $response"
           exit 1
         fi
+        sleep 5 # we make sure that the workflow is created before checking if it exists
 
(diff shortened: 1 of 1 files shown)


Ronan-Gorain (approved):

mixins/elasticsearch.yml
```diff
@@ -78,6 +78,7 @@ actions:
           echo "Pipeline trigger failed: $response"
           exit 1
         fi
+        sleep 5 # we make sure that the workflow is created before checking if it exists
```
kevinvandel, line 81: @adoux is that really necessary ? You tried without an ha an issue ?

adoux: I didn't try, just my instinct that it could take sometimes 1 or 2 seconds for the workflow to be created if CircleCi is struggling or anything else, so just to make sure I added this sleep

PR #314, opened 2022-04-12
PAAS-2019: fail if no workflow only once

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2019

Short description:

### DIFF
--- a/mixins/elasticsearch.yml
+++ b/mixins/elasticsearch.yml
@@ -78,11 +78,15 @@ actions:
           echo "Pipeline trigger failed: $response"
           exit 1
         fi
-        sleep 5 # we make sure that the workflow is created before checking if it exists
 
         # Skip the timeout while updating a mutualized EC deployment but wait till end if it's a new EC deployment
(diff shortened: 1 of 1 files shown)


mixins/elasticsearch.yml
```diff
@@ -78,11 +78,15 @@ actions:
           echo "Pipeline trigger failed: $response"
           exit 1
         fi
-        sleep 5 # we make sure that the workflow is created before checking if it exists
 
         # Skip the timeout while updating a mutualized EC deployment but wait till end if it's a new EC deployment
         if [ "$ec_deployment_mutualized" = "false" ] || [ "$ec_deployment_es_endpoint" = null ]; then
           pipelineId=$(echo $response | jq -r .id)
+          response=$(curl -LSs "https://circleci.com/api/v2/pipeline/$pipelineId/workflow" -H "Circle-Token: $CCToken")
+          if [ "$(echo $response | jq -r '.items[0].status')" == "null" ]; then
```
Ronan-Gorain, line 86: I thought we decided to add this if block in order to set the old condition back into the loop

Ronan-Gorain (changes requested):

Ronan-Gorain (approved):

PR #316, opened 2022-04-22
PAAS-2029 fix indices without analyser

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-2029

Short description:

### DIFF
--- a/packages/one-shot/paas-2029-fix-analyser-missing-in-indices-after-ec-migration.yml
+++ b/packages/one-shot/paas-2029-fix-analyser-missing-in-indices-after-ec-migration.yml
@@ -1,0 +1,182 @@
+---
+type: update
+version: 1.5.2
+name: Jahia - PAAS-2029 fix indices without analyser after jCustomer EC preECMigration
+id: jahia-PAAS-2029-fix-indices-without-analyser-after-jCustomer-EC-migration
+description:
+  short: Jahia - PAAS-2029 fix indices without analyser after jCustomer EC migration
+
+mixins:
+  - "../../mixins/common.yml"
+  - "../../mixins/jcustomer.yml"
+
+onInstall:
+  - isReindexNeeded
+  - if(${response.out}):
+      - manageFRO:
+          enable: true
+      - muteDatadogHost:
+          target: "cp"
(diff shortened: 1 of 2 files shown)


packages/one-shot/paas-2029-fix-analyser-missing-in-indices-after-ec-migration.yml
```diff
@@ -0,0 +1,182 @@
...
+            else
+              echo "true"
+              exit 0
+          fi
+        done
+        echo "false"
+
+  reindex:
+    - cmd[${nodes.cp.first.id}]: |-
+        output_file="/tmp/paas-2029.log"
```
Ronan-Gorain, line 88: Wouldn't it be better to use files indexed by datadog here ? I'm thinking to /var/log/jelastic-packages  folder.

packages/one-shot/paas-2029-fix-analyser-missing-in-indices-after-ec-migration.yml
```diff
@@ -0,0 +1,182 @@
...
+      - unmuteDatadogHost:
+          target: "cp"
+  - else:
+      return:
+        type: success
+        message: "Indices are ok"
+
+
+actions:
+  controlDaemon:
```
Ronan-Gorain, line 37: This will shutdown all nodes at the same time. Don't we prefer a rolling restart here ?

laurentfufu: I imagine that rolling restart of jcustomer does not bring anything a priori as we recreate the indexes on elasticsearch. This was also done during the migration to EC (and advised by the product), so I stick with this.

packages/one-shot/paas-2029-fix-analyser-missing-in-indices-after-ec-migration.yml
laurentfufu: I'm not sure what the point of having these debug logs from a one-shot manifest in datadog is but why not. Confirm and I will do :stuck_out_tongue:

PR #324, opened 2022-04-27
Paas 2026 clean revisions table

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2026

Short description:


mixins/jahia.yml
```diff
@@ -876,6 +876,35 @@ actions:
...
+    - cmd [proc]: |-
+        batch_size=${this.batchSize}
+        sql_cmd="mysql -u $DB_USER -p$DB_PASSWORD -h 127.0.0.1 -D jahia -P6033"
+        query="SELECT MIN(REVISION_ID) FROM JR_J_LOCAL_REVISIONS;"
+        min_revision_number=$($sql_cmd -sN -e "$query")
+        if [ $? -ne 0 ]; then
+          echo "Can't get the minimum revision number, aborting"
+          exit 1
+        fi
+        query="DELETE FROM JR_J_JOURNAL where REVISION_ID < $min_revision_number limit 100000"
```
kevinvandel, line 891: You should use `$batch_size` instead of hard-coding 100000 in the query.
Also, are we sure this will delete rows in the "correct order" ? (lower IDs first)

kevinvandel (changes requested):

mixins/jahia.yml
laurentfufu: about the _correct order_, an `ORDER BY` should be added

laurentfufu (approved):

PR #334, opened 2022-05-03
PAAS-1786: Check that module state did not change during PAT and SAM

Ronan-Gorain:

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v14.yaml
+++ b/packages/jahia/migrations/migrate-to-v14.yaml
@@ -68,7 +68,8 @@ actions:
       user: root
 
     - getJahiaVersion
-    - enableKarafLogin: proc
+    - enableKarafLogin: "cp, proc"
+    - saveModulesState
     - removeHealthcheckModule
 
     - isVersionHigherOrEqual:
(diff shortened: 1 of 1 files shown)


packages/jahia/migrations/migrate-to-v14.yaml
```diff
@@ -291,3 +310,53 @@ actions:
             fi
           fi
           echo "[ERROR] Healthcheck result different from GREEN or YELLOW, exiting" 1>&2 && exit 1
+    - if ("${globals.healthCheckState}" == "Up-to-Date"):
+        - checkJahiaHealth: "cp, proc"
+
+  saveModulesState:
+    - cmd[cp,proc]: |-
+        # Don't know why but running this command in one line (like the next cmd block) failed hafl of the time.
```
laurentfufu, line 318:
```suggestion
        # Don't know why but running this command in one line (like the next cmd block) failed half of the time.
```

Ronan-Gorain: fixed

kevinvandel (approved):

PR #341, opened 2022-05-05
Paas 1786

Ronan-Gorain:

Short description:

### DIFF
--- a/packages/jahia/install.yml
+++ b/packages/jahia/install.yml
@@ -289,6 +289,7 @@ onInstall:
   - setCloudConf
   - install:
       jps: "${baseUrl}/update-events.yml"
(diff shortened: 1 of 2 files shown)


packages/jahia/install.yml
```diff
@@ -290,6 +290,7 @@ onInstall:
   - install:
       jps: "${baseUrl}/update-events.yml"
   - startupJahiaHealthCheck: cp, proc
+  - checkPatGroovyScriptExecution
```
kevinvandel, line 293: Shouldn't this be done before `startupJahiaHealthCheck` ? Since this action requires the PAT token.

kevinvandel (changes requested):

kevinvandel (approved):

PR #348, opened 2022-05-11
PAAS-1543 Re-enable io check in datadog agent

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1543

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v15.yaml
+++ b/packages/jahia/migrations/migrate-to-v15.yaml
@@ -1,0 +1,80 @@
+---
+type: update
+version: 1.5.2
+name: Migrate Jahia env to v15
+id: migrate-jahia-env-v15
+
+# Required for healthchecks
+mixins:
+  - "../../../mixins/common.yml"
+  - "../../../mixins/jahia.yml"
+  - "../../../mixins/mariadb.yml"
+  - "../../../mixins/haproxy.yml"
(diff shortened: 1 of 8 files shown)


packages/jahia/migrations/migrate-to-v15.yaml
```diff
@@ -0,0 +1,80 @@
...
+  - checkMariadbDatadogCustomChecks: sqldb
+  - eventsUpdate
+  - setGlobalRepoRootUrl
+  ### End Pre-checks
+
+  # Actions that require a restart:
+  - enableIoCheck                                         # PAAS-1543
+
+  # Restart actions:
+  - restartDDAgentOnJahiaNodes                            # needed for PAAS-1543
```
kevinvandel, line 37: Why split in 2 actions ?

Ronan-Gorain (approved):

packages/jahia/migrations/migrate-to-v15.yaml
laurentfufu: I just went back to the way it was done when the check was disable, but I can change that, no worries. You just have to confirm it :)

kevinvandel: ok so that's because it's a copy-paste, that explains it
no worries, I just wanted to understand

PR #351, opened 2022-05-16
PAAS-2044: Add datadog APM opts to fix forms issue

amitc-jahia:
JIRA issue: https://jira.jahia.org/browse/PAAS-2044

Short description:


packages/jahia/configure-apm.yaml
```diff
@@ -12,6 +12,18 @@ mixins:
 onInstall:
   - env.control.AddContainerEnvVars[cp, proc]:
       vars: {"DATADOG_APM_ENABLED": "${settings.apmEnabled}"}
+  - if(settings.apmEnabled):
```
Ronan-Gorain, line 15: APM configuration is already handled there https://github.com/Jahia/jelastic-packages/blob/6d7df88ae616c7f68e47a1d00d6e3519ef598374/mixins/jahia.yml#L193
I think it's better to add the options to APT_OPTS directly there

amitc-jahia: Yes it could be added here directly. But I added the options while enabling the APM on an env as per mentioned in the ticket description. So having the options when APM is not enabled shouldn't be an issue (I guess)?

amitc-jahia: ok :+1:

amitc-jahia: Fixed

adoux (approved):

PR #355, opened 2022-05-20
PAAS-1973: Fix duplication of lines in tomcat env file

amitc-jahia:
JIRA issue: https://jira.jahia.org/browse/PAAS-1973

Short description:
Please check ticket comments


packages/jahia/migrations/migrate-to-v15.yaml
```diff
@@ -109,3 +111,9 @@ actions:
           vars:
             jahia_cfg_mvnPath: ${globals.jahia_cfg_mvnPath}
         - jahiaRollingRestart
+
+  deleteDuplicateTomcatEnvLines:
+    cmd[cp, proc]: |-
+      tomcat_env=/opt/tomcat/conf/tomcat-env.sh
+      sed -i '0,/^APM_OPTS/b;/^APM_OPTS/d' ${tomcat_env}
+      sed -i '0,/^$DATADOG_APM_ENABLED/b;/^$DATADOG_APM_ENABLED/d' ${tomcat_env}
```
laurentfufu, line 119: I'm not sure it's a good idea to keep the first line found because we might reuse an old configuration.

laurentfufu (changes requested):

packages/jahia/migrations/migrate-to-v15.yaml
amitc-jahia: Fixed

packages/jahia/migrations/migrate-to-v15.yaml
```diff
@@ -109,3 +111,23 @@ actions:
...
+    cmd[cp, proc]: |-
+      tomcat_env=/opt/tomcat/conf/tomcat-env.sh
+      echo "JMX_OPTS
+      jahia_cfg_cluster_tcp_bindAddress
+      APM_OPTS
+      \$DATADOG_APM_ENABLED" > /tmp/lines_to_check
+      tac ${tomcat_env} > /tmp/reverse_tomcat-env.sh
+      while read line; do
+        sed -i "0,/^$line/b;/^$line/d" /tmp/reverse_tomcat-env.sh
+      done < /tmp/lines_to_check
```
laurentfufu, line 125: I think an array like that would have been more readable here but ok:
```suggestion
      lines_to_checks=(
        "JMX_OPTS"
        "jahia_cfg_cluster_tcp_bindAddress"
        "APM_OPTS"
        "\$DATADOG_APM_ENABLED"
      )
      tac ${tomcat_env} > /tmp/reverse_tomcat-env.sh
      for line in ${lines_to_checks[*]}; do
        sed -i "0,/^$line/b;/^$line/d" /tmp/reverse_tomcat-env.sh
      done
```

laurentfufu (approved):

PR #358, opened 2022-05-25
PAAS-2070 Use CRLF for haproxy's httpchk headers

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-2070

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v15.yaml
+++ b/packages/jahia/migrations/migrate-to-v15.yaml
@@ -36,6 +36,7 @@ onInstall:
   - enableIoCheck                                         # PAAS-1543
   - addAPMOptsForFormsIssue                               # PAAS-2044
   - updateMavenPathEnvVar                                 # PAAS-2038
+  - haproxyCheckConformRFC                                # PAAS-2070
 
   # Restart actions:
   - restartDDAgentOnJahiaNodes                            # needed for PAAS-1543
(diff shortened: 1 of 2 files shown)


packages/jahia/migrations/migrate-to-v15.yaml
```diff
@@ -34,6 +34,7 @@ onInstall:
   - enableIoCheck                                         # PAAS-1543
   - addAPMOptsForFormsIssue                               # PAAS-2044
   - updateMavenPathEnvVar                                 # PAAS-2038
+  - haproxyCheckConformRFC                                # COPS-571
```
kevinvandel, line 37: Wrong ticket ID :stuck_out_tongue:

kevinvandel (changes requested):

packages/jahia/migrations/migrate-to-v15.yaml
laurentfufu: fixed :trollface:

packages/jahia/migrations/migrate-to-v15.yaml
```diff
@@ -109,3 +110,18 @@ actions:
...
+    - cmd[bl]: |-
+        HAPROXY_CONF_DIR=/etc/haproxy/haproxy.cfg.jahia
+        hash=$(grep -E "\s+option httpchk GET" $HAPROXY_CONF_DIR/jahia-cloud.cfg | md5sum | awk '{print $1}')
+        if [[ "$hash" == "15b46715eca4f35b86c62579986789fa" ]]; then
+          echo "up to date"
+        else
+          # fetches the new haproxy configuration conforming to RFC 7320 (aka "always use CRLF")
+          stats_pwd=$(awk '$1=="password:" {print $2}' /etc/datadog-agent/conf.d/haproxy.d/conf.yaml)
+          curl -fLSso $HAPROXY_CONF_DIR/jahia-cloud.cfg ${globals.repoRootUrl}/assets/haproxy/haproxy-jahia-cloud.cfg || exit 1
+          sed -i "s/admin-stat-password/$stats_pwd/g" /etc/haproxy/haproxy.cfg.jahia/jahia-cloud.cfg
```
Ronan-Gorain, line 124: I might be wrong, but it looks like the case where the auth basic is enabled is not handled (including  user env-admin password  )

Ronan-Gorain (dismissed):

packages/jahia/migrations/migrate-to-v15.yaml
laurentfufu: should be better now :trollface:

kevinvandel (approved):

kevinvandel (approved):

PR #362, opened 2022-06-01
PAAS-1916: Use Jahia API to handle actions on modules

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1916

Short description:
* The install/uninstall/start API requests are run on the processing only because they are applied to the whole cluster
* I tried to make actions as generic as possible so they can be reused for other needs
* I made sure the `install-module.yml` wasn't used anywhere and renamed/updated it to be used to uninstall jExperience on linked Jahia envs when deleting a jCustomer env


mixins/jahia.yml
```diff
@@ -520,114 +518,321 @@ actions:
...
+    #   - globals.moduleStarted: true if the module is started, false otherwise
+    - setGlobals:
+        moduleStarted: true
+    - set:
+        curl: 'curl -SsH "Authorization: APIToken ${globals.__secret__pat_token}"'
+    - getPatTokenAndKey
+    - cmd [proc]: |-
+        module_info=$(${this.curl} "localhost/modules/api/bundles/${this.module_name}/${this.module_version}/_localInfo")
+        case "$(jq -r .moduleState <<< $module_info)" in
+          STARTED)
```
kevinvandel, line 708: I'm 99% sure we'll have issues with this. For instance I know that in some cases the module status is "INSTALLED", while it's actually started... I'll try to reproduce this.

mixins/jahia.yml
```diff
@@ -520,114 +518,321 @@ actions:
...
+            echo "Module started"
+            ;;
+          ""|null|UNINSTALLED) # If 401 unauthorized or if the module is not installed, it should fail
+            echo "401 unauthorized"
```
kevinvandel, line 712: generic message, you know the drill

mixins/jahia.yml
```diff
@@ -520,114 +518,321 @@ actions:
...
+        - log: "Versions to delete for ${this.module_name} module: ${response.out}"
+        - script: |-
+            return {"result": 0, "versions": "${response.out}".split(";")}
+        - forEach(response.versions):
+            - uninstallModuleVersion:
+                module_name: ${this.module_name}
+                module_version: ${@i}
+        - log: Old versions of ${this.module_name} module uninstalled
+
+  uninstallModuleCompletely:
```
kevinvandel, line 800: Why not just `uninstallModule` ?

mixins/jahia.yml
```diff
@@ -520,114 +518,321 @@ actions:
...
+    - getPatTokenAndKey
+    - log: Uninstalling all versions of ${this.module_name} module on ${this.target}
+    - cmd[proc]: |-
+        resp=$(${this.curl} "localhost/modules/api/bundles/${this.module_name}/*/_localInfo")
+        case "$(jq -r ". | length" <<< $resp)" in
+          "0") # Means the module is not installed
+            echo "not installed"
+            ;;
+          "") # 401 unauthorized so no JSON resp
+            echo "401 unauthorized"
```
kevinvandel, line 814: you know what

mixins/jahia.yml
```diff
@@ -520,114 +518,321 @@ actions:
...
-    - performModuleUpgradeOnNode: proc
-    - foreach (nodes.cp):
-        - performModuleUpgradeOnNode: ${@i.id}
+  upgradeModule:
+    # Parameters:
+    #   - module_name: the module to upgrade
+    #   - module_version: the version of the module to install
+    - checkIfModuleVersionIsInstalled:
+        module_name: ${this.module_name}
+        module_version: ${this.module_version}
```
kevinvandel, line 527: That may but out-of-scope, but I'm wondering if we shouldn't also make sure that the module isn't installed in a version above the one we are about to install, to avoid downgrading it if that's the case.

packages/common/restore.yml
```diff
@@ -312,28 +314,17 @@ actions:
         - cmd[proc]: |-
             echo "Upgrade jexperience module to initial v${globals.jexperience_active_version} now" >> ${globals.logsPath}
         - log: "Install jExperience v${globals.jexperience_active_version}"
-        - upgradeJahiaModule:
-            module: jExperience
+        - upgradeModule:
+            module: jexperience
```
kevinvandel, line 318: Parameters name seem wrong (should be `module_name` and `module_version` from what I see)

mixins/jahia.yml
```diff
@@ -520,114 +518,321 @@ actions:
...
-        moduleVersion: "${this.version}"
-        awkFS: "-F ' ?*\t ?'"
-        awkIC: "-v IGNORECASE=1"
-
-    - performModuleUpgradeOnNode: proc
-    - foreach (nodes.cp):
-        - performModuleUpgradeOnNode: ${@i.id}
+  upgradeModule:
+    # Parameters:
+    #   - module_name: the module to upgrade
```
kevinvandel, line 523: From what I understand, the `module_name` must be what seems to be called the "symbolic name", so maybe rename it to `module_symname` to make things clear? Or at least updated the comment.

packages/one-shot/ccm-103-update-to-tomcat-9-0-54.yaml
```diff
@@ -18,7 +18,7 @@ actions:
   updateJahiaCsrfGuardModule:
     - getJahiaVersion
     - if( "${globals.jahiaVersion}" == "7.3.7.0" || "${globals.jahiaVersion}" == "7.3.8.0" ):
-        - upgradeJahiaModule:
+        - upgradeModule:
```
kevinvandel, line 21: Parameters name must be updated too.

packages/jahia/link-to-jcustomer.yml
```diff
@@ -25,8 +25,8 @@ onInstall:
   - getJexperienceVersion
   - refreshUnomiAllowedIPs:
       unomiEnvName: ${settings.unomienv}
-  - upgradeJahiaModule:
-      module: jExperience
+  - upgradeModule:
```
kevinvandel, line 28: Parameters name must be updated too.

packages/jahia/manage-module.yml
```diff
@@ -0,0 +1,41 @@
...
+
+mixins:
+  - ../../mixins/common.yml
+  - ../../mixins/jahia.yml
+
+onInstall:
+  - if ("${settings.action}" == "install"):
+    - upgradeModule:
+        module: ${settings.moduleName}
+        version: ${settings.moduleVersion}
```
kevinvandel, line 17: Wrong parameters name.

packages/jahia/migrations/migrate-to-v11.yaml
```diff
@@ -89,7 +89,7 @@ actions:
             moduleId: ${response.out}
         - disableKarafLogin: proc
         - if (${this.moduleId}):
-            - upgradeJahiaModule:
+            - upgradeModule:
```
kevinvandel, line 92: Parameters name must be updated too.

packages/jahia/migrations/migrate-to-v9.yaml
```diff
@@ -181,7 +181,7 @@ actions:
     - else:
         - set:
             ds_version: 3.1.0
-    - upgradeJahiaModule:
+    - upgradeModule:
```
kevinvandel, line 184: Parameters name must be updated too.

mixins/jahia.yml
kevinvandel: Note that this comment applies to other actions too, but I'm too lazy to copy my comment everywhere :p

kevinvandel (changes requested):

mixins/jahia.yml
kevinvandel: OK so I only reproduce this behaviour with the `cluster:bundle-list` command, so I guess we'll be good.
```
$ curl -s -H "authorization: APIToken $token" http://localhost:8080/modules/api/bundles/org.jahia.modules/distributed-sessions/*/_info | jq
{
  "browsing.45955": {
    "org.jahia.modules/distributed-sessions/3.2.1": {
      "type": "MODULE",
      "osgiState": "ACTIVE",
      "moduleState": "STARTED"
    }
  },
  "processing.45954": {
    "org.jahia.modules/distributed-sessions/3.2.1": {
      "type": "MODULE",
      "osgiState": "ACTIVE",
      "moduleState": "STARTED"
    }
  }
}
```
```
fraise@dx()> bundle:list -s | grep dist
118 │ Active   │  90 │ 3.2.1           │ distributed-sessions
fraise@dx()> cluster:bundle-list -s default | grep dist
118 │ Installed │     │ cluster/local │         │ 3.2.1                 │ distributed-sessions
```

mixins/jahia.yml
```diff
@@ -520,114 +518,321 @@ actions:
...
+        module_version: ${this.module_version}
+    - if (!${globals.moduleStarted}):
+        - startModule:
+            module_name: ${this.module_name}
+            module_version: ${this.module_version}
+    - uninstallOldVersionsOfModule:
+        module_name: ${this.module_name}
+        module_version: ${this.module_version}
+
+  checkIfModuleIsInstalled:
```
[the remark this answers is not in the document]
adoux: Nope, it's gonna be used in PAAS-2061

mixins/jahia.yml
```diff
@@ -520,114 +518,321 @@ actions:
...
+        curl: 'curl -SsH "Authorization: APIToken ${globals.__secret__pat_token}"'
+    - getPatTokenAndKey
+    - cmd [proc]: |-
+        resp=$(${this.curl} "localhost/modules/api/bundles/${this.module_name}/${this.module_version}/_localInfo")
+        case "$(jq -r .moduleState <<< $resp)" in
+          INSTALLED|STARTED|RESOLVED|STARTING|STOPPING|ACTIVE) # If the module is not installed, the value would be "UNINSTALLED" or "null" if 404 is returned
+            echo "Module installed"
+            ;;
+          "") # If it's not JSON because of 401 unauthorized (wrong token for instance)
+            echo "401 unauthorized"
```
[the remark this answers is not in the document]
adoux: Fixed

mixins/jahia.yml
```diff
@@ -520,114 +518,321 @@ actions:
...
+            ;;
+          *) # Otherwise it means the module was installed, nothing to do
+            ;;
+        esac
+        rm -rf $tmp_dir
+    - if ("${response.out}" == "error"):
+        - return:
+            type: error
+            message: "An error occurred during module installation."
+    - log: Module ${this.module_name} v${this.module_version} has been installed on ${this.target}
```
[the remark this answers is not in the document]
adoux: Oops, oversight

packages/one-shot/ccm-103-update-to-tomcat-9-0-54.yaml
adoux: Oversight because old package but it's better to update it indeed. Fixed

packages/jahia/manage-module.yml
adoux: Oversight, fixed

packages/jahia/migrations/migrate-to-v11.yaml
adoux: Oversight because old package but it's better to update it indeed. Fixed

packages/jahia/migrations/migrate-to-v9.yaml
adoux: Oversight because old package but it's better to update it indeed. Fixed

mixins/jahia.yml
```diff
@@ -520,114 +518,340 @@ actions:
...
-    - foreach (nodes.cp):
-        - performModuleUpgradeOnNode: ${@i.id}
+  upgradeModule:
+    # Installs the specified version of a module and deletes all the previous versions if installed.
+    #
+    # Parameters:
+    #   - module_symname: the symbolic name of the module to upgrade (for instance "distributed-sessions")
+    #   - module_version: the version of the module to install
+    - checkIfModuleVersionIsInstalled:
+        module_symname: ${this.module_name}
```
kevinvandel, line 528: You forgot to rename to module_symname here and below

mixins/jahia.yml
kevinvandel: I'm still waiting for an answer :stuck_out_tongue:

mixins/jahia.yml
kevinvandel: @adoux didn't you tell me you would do it ?

mixins/jahia.yml
adoux: True, makes more sense. Fixed!

mixins/jahia.yml
adoux: I did, but forgot :man_shrugging: 
Did it in the `checkIfModuleIsInstalled` action, which makes more sense to me

kevinvandel: There are two issues:
- in the case we don't install the target module version because a more recent version is installed, `upgradeModule` will fail at `checkIfModuleVersionIsInstalled` and `checkIfModuleVersionIsStarted` steps (if I'm not mistaken)
- you check the "most recent installed version", but we would also need to check it is running ; in the end what we need is the "latest _running_ version"

mixins/jahia.yml
```diff
@@ -520,114 +518,356 @@ actions:
...
+        if [[ "${this.jar_url}" == "" ]]; then
+          url="${this.modulesUrl}/${this.module_symname}/${this.module_version}/${this.module_symname}-${this.module_version}.jar"
+        else
+          url=${this.jar_url}
+        fi
+        tmp_dir=$(mktemp -d)
+        local_jar_file=$tmp_dir/${this.module_symname}-${this.module_version}.jar
+        curl -fLSso $local_jar_file $url >&2
+        if [ $? -ne 0 ]; then
+          echo "download failed"
```
laurentfufu, line 662: if _download failed_ the action will stop here (with success) because the case isn't handle with an `if ("${response.out}" == "download failed")` like _error_ output is.

laurentfufu (changes requested):
looks nice but I don't understand the point of modifying the v9 and v11 migration manifests :grey_question:

mixins/jahia.yml
```diff
@@ -520,114 +518,272 @@ actions:
...
-
-  performModuleUpgradeOnNode:
-    - enableKarafLogin: ${this}
-    - installModuleIfNotPresent: ${this}
-    - removeOldModuleVersions: ${this}
-    - disableKarafLogin: ${this}
-
-  installModuleIfNotPresent:
+  upgradeModule:
+    # Installs the specified version of a module and deletes all the previous versions if installed.
```
kevinvandel, line 522: "deletes all the previous versions if installed" isn't really true (in some cases we may uninstall more recent versions)
Also I think that adding some detail about the various cases (module is already installed but not running, or running in a newer version than the target one) wouldn't hurt

kevinvandel (changes requested):

mixins/jahia.yml
adoux: Indeed, it's been improved now

mixins/jahia.yml
adoux: I updated comments, added new ones, and improved the upgradeModule action so we make sure we are running the right version before uninstalling the other ones

kevinvandel (dismissed):

Ronan-Gorain:
I have one concern about the checkModule action. I think it's not possible to know is a specific version is installed or not. So if I want to install a specific version of the module but there is already another version installed and running, I have no way to know if the version I want to install is already installed if it's not running.

adoux:
> I have one concern about the checkModule action. I think it's not possible to know is a specific version is installed or not. So if I want to install a specific version of the module but there is already another version installed and running, I have no way to know if the version I want to install is already installed if it's not running.

I added a `checkVersionOfModule` action and improved the `installModule` action to not reinstall the module if the version is already installed/running

Ronan-Gorain (approved):

PR #368, opened 2022-06-13
PAAS-1916: fix if condition checking module state in checkModuleAction

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-1916

Short description:

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -580,7 +580,7 @@ actions:
     # Parameters:
     # - moduleSymname: the symbolic name of the module to upgrade (for instance "distributed-sessions")
     # Returns:
-    # - globals.moduleState: 
+    # - globals.moduleState:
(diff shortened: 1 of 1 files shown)


mixins/jahia.yml
```diff
@@ -617,7 +617,7 @@ actions:
         - setGlobals:
             moduleState: ${response.out}
     - if ("${globals.moduleState}" == "started") || ("${globals.moduleState}" == "installed"):
-        - if ("${response.out}" == "module started"):
+        - if ("${globals.moduleState}" == "started"):
```
kevinvandel, line 620: This should be converted into a `if... elif...` statement

kevinvandel (changes requested):

kevinvandel (approved):

PR #369, opened 2022-06-13
PAAS-2061: upgrade jexperience for 7.3 and 8.1

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2061

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v15.yaml
+++ b/packages/jahia/migrations/migrate-to-v15.yaml
@@ -32,6 +32,7 @@ onInstall:
   ### End Pre-checks
 
   - deleteDuplicateTomcatEnvLines                         # PAAS-1973
+  - upgradeJexperience                                    # PAAS-2061
 
   # Actions that require a restart:
   - enableIoCheck                                         # PAAS-1543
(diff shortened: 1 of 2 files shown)


packages/jahia/migrations/migrate-to-v15.yaml
```diff
@@ -193,3 +194,31 @@ actions:
...
+    # Update jExperience to 1.11.8 on Jahia 7.3 envs, and 2.2.2 on 8.1 envs
+    - getJahiaVersion
+    - isVersionStrictlyLower:
+        a: ${globals.jahiaVersion}
+        b: 8.0.0.0
+        res: isLowerThan8000
+    - if( ${globals.isLowerThan8000} ):
+        - checkModule:
+            moduleSymname: jexperience
+        - if ("${globals.moduleState}" == "started"):
```
kevinvandel, line 208: As discussed, instead of checking module status, it would be better to check if the env is linked to a jCustomer env (i.e. if envLink is defined). This check can be done at the beginning of the action, even before `getJahiaVersion`.

kevinvandel (changes requested):

packages/jahia/migrations/migrate-to-v15.yaml
adoux: Done!

kevinvandel (approved):

PR #379, opened 2022-06-20
PAAS-1855 add AS installation manifest

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1855

Short description:


packages/jahia/augmented-search-install.yml
```diff
@@ -0,0 +1,218 @@
...
+          echo "Failed to use the $es_connection_name connection for AS"
+          exit 1
+        fi
+        echo "AS is now using the new ES connection"
+
+  setupModules:
+    - setupAugSearch
+    - setupESConnector
+
+# settings:
```
Ronan-Gorain, line 212: To be remove :)

Ronan-Gorain (changes requested):

packages/jahia/augmented-search-install.yml
laurentfufu: oops :grin: 
fixed

Ronan-Gorain (approved):

PR #382, opened 2022-06-21
PAAS-2001: handle muting of envname based monitors at

Ronan-Gorain:
stop/start/delete/restore/upgrade

JIRA issue: https://jira.jahia.org/browse/PAAS-2001

Short description:

### DIFF
--- a/packages/common/restore.yml
+++ b/packages/common/restore.yml
@@ -20,6 +20,8 @@ onInstall:
   - muteDatadogHost:
       target: "*"
       duration: 240 # 4h
+  - muteEnvWideMonitors:
+      duration: 4 # 4h
   - setGlobalRepoRootUrl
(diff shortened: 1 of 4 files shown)


packages/common/restore.yml
```diff
@@ -20,6 +20,8 @@ onInstall:
   - muteDatadogHost:
       target: "*"
       duration: 240 # 4h
+  - muteEnvWideMonitors:
+      duration: 4 # 4h
```
kevinvandel, line 24: I think it's highly confusing and error-prone to have the `muteDatadogHost` duration in minutes and the `muteEnvWideMonitors` one in hours...

Ronan-Gorain: updated

adoux (approved):

PR #391, opened 2022-06-27
PAAS-2057: Enable Datadog APM by default 

kevinvandel:
JIRA issue: https://jira.jahia.org/browse/PAAS-2057


packages/jahia/migrations/migrate-to-v16.yaml
```diff
@@ -30,17 +30,12 @@ onInstall:
...
   ### End Pre-checks
 
-  # Actions that require a restart:
-  - removeModulesFolderAndRestartAllJahia                     # PAAS-2079
   - upgradeSam                                                # PAAS-2089
 
   # Actions that require a restart:
   - setESPrefixNodeGroupData                                  # PAAS-1855
-
-  # Restart action:
```
Ronan-Gorain, line 40: If you remove this, jahia, I fear 7 environments won't be restarted because the removeModulesFolderAndRestartAllJahia action only take care of jahia 8 environments

Ronan-Gorain (changes requested):

packages/jahia/migrations/migrate-to-v16.yaml
kevinvandel: Indeed, I didn't see the condition on Jahia 8. I'll rework that.

kevinvandel: Fixed !

adoux (approved):

PR #395, opened 2022-06-29
PAAS-1960: Handle Augmented search module at backup restore

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-1960

Short description:

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -767,11 +767,42 @@ actions:
     - else:
         - log: ${this.moduleSymname}/${this.moduleVersion} is already running, nothing to do
 
+  stopModule:
+    # Stops the running version of a module.
+    #
+    # Parameters:
+    #   - moduleSymname: the symbolic name of the module to stop (for instance "distributed-sessions")
+    - set:
+        curl: 'curl -SsH "Authorization: APIToken ${globals.__secret__pat_token}"'
+    - getPatTokenAndKey
+    - log: Stopping ${this.moduleSymname}
+    - cmd [proc]: |-
+        resp=$(${this.curl} -XPOST -d "" \
+                 "localhost/modules/api/bundles/${this.moduleSymname}/_stop")
+        case "$(jq -r .bundleInfos <<< $resp)" in
+          "null") # If stop failed, there is no bundleInfos attribute in the JSON response
+            echo "error"
+            jq -r .message <<< $resp >&2
+            ;;
+          "") # if response is not JSON, for instance 401 unauthorized
+            echo "error"
+            echo $resp >&2
+            ;;
+          *) # Otherwise it means the module was stopped, nothing to do
+            ;;
+        esac
+    - if ("${response.out}" == "error"):
+        - return:
+            type: error
+            message: "An error occurred during module stop."
+    - log: ${this.moduleSymname}/${this.moduleVersion} has been stopped
+
+
   startModule:
     # Starts the specified version of a module.
     #
     # Parameters:
-    #   - moduleSymname: the symbolic name of the module to upgrade (for instance "distributed-sessions")
+    #   - moduleSymname: the symbolic name of the module to start (for instance "distributed-sessions")
     #   - moduleVersion: the version of the module to install
     - set:
         curl: 'curl -SsH "Authorization: APIToken ${globals.__secret__pat_token}"'
@@ -789,13 +820,13 @@ actions:
             echo "error"
             echo $resp >&2
             ;;
-          *) # Otherwise it means the module was installed, nothing to do
+          *) # Otherwise it means the module was started, nothing to do
             ;;
         esac
     - if ("${response.out}" == "error"):
         - return:
             type: error
-            message: "An error occurred during module uninstall."
+            message: "An error occurred during module startup."
     - log: ${this.moduleSymname}/${this.moduleVersion} has been started
 
   uninstallModuleVersion:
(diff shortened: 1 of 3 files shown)


packages/common/restore.yml
```diff
@@ -313,13 +342,51 @@ actions:
...
+            elasticsearch_password: ${fn.password(20)}
+        - createESAccount4AS
+        - removeDefaultESConnection
+        - setDefaultESConnection
+        - setAugSearchESConnection:
+        - triggerAugSearchFullReindex
+        - if ("${this.as-state}" != "started"):
+            # Stops AS module if it was not running on the env before the restore
+            stopModule:
+              moduleSymname: augmented-search
```
laurentfufu, line 383: if AS on the target env was initially present but not started, maybe we can avoid to trigger a full reindex just before stop the module ?

packages/common/restore.yml
```diff
@@ -313,13 +342,51 @@ actions:
...
+            as-state: ${response.out}
+        - if ("${globals.moduleState}" == "started")  && ("${this.as-version}" != "${globals.runningVersion}"):
+            # Stop module if AS backup and environment version mismatch
+            stopModule:
+              moduleSymname: augmented-search
+        - installOrUpgradeModule:
+            moduleSymname: augmented-search
+            moduleVersion: ${this.as-version}
+        - uninstallStoppedVersionsOfModule:
+            moduleSymname: augmented-search
```
laurentfufu, line 372: `installOrUpgradeModule` already use this action

mixins/jahia.yml
```diff
@@ -1316,6 +1387,112 @@ actions:
...
+        current_es_connection=$(curl -fLSs \
+          -H "Authorization: APIToken $__secret__API_TOKEN" \
+          -H 'Origin: http://localhost:8080' \
+          -H 'Content-Type: application/json' \
+          -d '{"query":"query { admin { search { currentConnection } } }"}' \
+          http://localhost:8080/modules/graphql | jq -r .data.admin.search.currentConnection)
+          if [ "$current_es_connection" = "$es_connection_name" ]; then
+            echo "Connection $es_connection_name already configured. Nothing to do"
+            exit 0
+          fi
```
laurentfufu, line 1481: `getAugSearchConnectionName` can already do that (and not used anywhere in this PR)

packages/common/restore.yml
Ronan-Gorain: removed

packages/common/restore.yml
Ronan-Gorain: You are right, I add a step to remove all indices before either trigger a reindexation or stopping the module.

mixins/jahia.yml
```diff
@@ -1316,6 +1387,129 @@ actions:
...
+        es_endpoint=$(echo $ec_deployment | jq -r .es_endpoint)
+
+        es_password=$(echo $ec_deployment_credentials | jq -r .password)
+        es_login=$(echo $ec_deployment_credentials | jq -r .login)
+
+        indices=$(curl "$es_endpoint/_cat/indices?format=json" -u "$es_login:$es_password")
+        indices_to_delete=($(echo $indices | jq --arg pattern "$jahia_cfg_elasticsearch_prefix_." -r '.[] | select(.index|test($pattern)) | .index'))
+        for index_to_delete in ${indices_to_delete[@]}; do
+          curl -X DELETE "$es_endpoint/$index_to_delete" -u "$es_login:$es_password"
+        done
```
laurentfufu, line 1510:
```suggestion
        curl -u "$es_login:$es_password" -XDELETE "$es_endpoint/${jahia_cfg_elasticsearch_prefix}_*"
```
Should do the trick (perhaps check that `jahia_cfg_elasticsearch_prefix` has the expected value as long as you do the delete with the elastic account).

At worst leave it like that but still fix the `jq --arg pattern "$jahia_cfg_elasticsearch_prefix_."` otherwise it will delete indices not related :trollface:

laurentfufu (changes requested):

mixins/jahia.yml
Ronan-Gorain: I added a check and fixed the jq pattern arg

Ronan-Gorain: I did not use ${jahia_cfg_elasticsearch_prefix} because it's interpreted by jelastic

mixins/jahia.yml
```diff
@@ -1316,6 +1387,130 @@ actions:
...
+            -d '{"query":"mutation { admin { search { setDbConnection(connectionId:\"jahia-cloud_augmented-search\") } } }"}' \
+            http://localhost:8080/modules/graphql)
+            result=$(echo $response | jq -r '.data.admin.search.setDbConnection')
+          if [ "$result" != "Successful" ]; then
+            echo "Failed to use the $es_connection_name connection for AS: $response"
+            exit 1
+          fi
+
+  purgeASIndicesInES:
+    # Set the jahia-cloud_augmented-search connexion from elasticsearch-connector module
```
laurentfufu, line 1487: Unrelated comment :trollface: 

Furthermore, this action could be agnostic and delete the env's indices whether they are for AS or jCustomer ?

Ronan-Gorain: Renamed in purgeEnvIndicesInES and moved to mixins/elasticsearch.yml
Updated patter from /${jahia_cfg_elasticsearch_prefix}__as_*  to /${jahia_cfg_elasticsearch_prefix}__*

laurentfufu (approved):

PR #399, opened 2022-06-30
PAAS-2022: fix default value for variable because of Cloudscripting

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2022

Short description:

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -234,10 +234,6 @@ actions:
         NODE_NAME=${HOSTNAME/-*}
         echo "hostname: ${_ROLE}.${NODE_NAME#node}" >> /etc/datadog-agent/datadog.yaml
 
-        # Enable APM by default
-        echo "apm_config:" >> /etc/datadog-agent/datadog.yaml
-        echo "  enabled: ${DATADOG_APM_ENABLED:-true}" >> /etc/datadog-agent/datadog.yaml
-
         sed 's/service: jahia/service: ${env.shortdomain}/' -i /etc/datadog-agent/conf.d/tomcat.d/conf.yaml
         sed 's/service: jahia/service: ${env.shortdomain}/' -i /etc/datadog-agent/conf.d/proxysql.d/conf.yaml
         sed 's/service: jahia/service: ${env.shortdomain}/' -i /etc/datadog-agent/conf.d/strongswan_connections_status.yaml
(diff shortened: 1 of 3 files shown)


mixins/jahia.yml
```diff
@@ -236,7 +236,8 @@ actions:
 
         # Enable APM by default
         echo "apm_config:" >> /etc/datadog-agent/datadog.yaml
-        echo "  enabled: ${DATADOG_APM_ENABLED:-true}" >> /etc/datadog-agent/datadog.yaml
+        [ -z ${DATADOG_APM_ENABLED} ] && DATADOG_APM_ENABLED=true
+        echo "  enabled: ${DATADOG_APM_ENABLED}" >> /etc/datadog-agent/datadog.yaml
```
laurentfufu, line 240: the main datadog conf file is handle by `set_dd_tags.sh` cronified  script and it's already put this line with `true` as default value (cf [here](https://github.com/Jahia/jahiastic-jahia/blob/master/base/files/set_dd_tags.sh#L13-L14))

mixins/jcustomer.yml
```diff
@@ -350,4 +350,5 @@ actions:
     - cmd [cp]: |-
         setenv=$(find /opt/jcustomer/jcustomer/bin -name setenv)
         echo 'APM_OPTS="-javaagent:'${JCUSTOMER_HOME}'/jcustomer/datadog/dd-java-agent.jar -Ddd.logs.injection=true -Ddd.service=jcustomer -Ddd.env='${env.envName}' -Ddd.profiling.enabled=true"' >> $setenv
-        echo '${DATADOG_APM_ENABLED:-true} && JAVA_OPTS+=" $APM_OPTS"' >> $setenv
+        echo '[ -z ${DATADOG_APM_ENABLED} ] && DATADOG_APM_ENABLED=true' >> $setenv
+        echo '${DATADOG_APM_ENABLED} && JAVA_OPTS+=" $APM_OPTS"' >> $setenv
```
laurentfufu, line 354:
```suggestion
        # this fucking cloudscripting interprets the bash form `${<variable name>:<something>}` as its own placeholder
        # so we have to trick it
        echo '${DATADOG_APM_ENABLED¤-true} && JAVA_OPTS+=" $APM_OPTS"' | sed 's/¤/:/' >> $setenv
```

I found this way more readable and has the advantage of explaining why a side trip is necessary here

mixins/jahia.yml
adoux: I removed this part

mixins/jcustomer.yml
adoux: Done, also done in the migration package

amitc-jahia (approved):

PR #418, opened 2022-07-07
fix bad test

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-1855

Short description:


kevinvandel (changes requested):
Ticket ID is missing in the commit message

kevinvandel (approved):

PR #424, opened 2022-07-14
PAAS-1879: update haproxy to v2.6.1

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1879

Short description:


mixins/haproxy.yml
```diff
@@ -21,21 +21,21 @@ actions:
     - setGlobalRepoRootUrl
     - cmd[${this}]: |-
         HAPROXY_CONF_DIR=/etc/haproxy/haproxy.cfg.jahia
-        HAPROXY_SYSCONFIG=/etc/sysconfig/haproxy
+        HAPROXY_PRESTART_SCRIPT=/usr/lib/systemd/scripts/haproxy_prestart.sh
```
laurentfufu, line 24: n my opinion the script should be in `/usr/local/bin` or even (maybe better) `/usr/local/sbin`.

assets/haproxy/haproxy_service_override
```diff
@@ -0,0 +1,6 @@
+[Service]
+ExecStartPre=/usr/lib/systemd/scripts/haproxy_prestart.sh
+ExecReload=/usr/lib/systemd/scripts/haproxy_prestart.sh
+ExecReload=/usr/local/sbin/haproxy -Ws -f $CONFIG -c -q $EXTRAOPTS
+ExecReload=/bin/kill -USR2 $MAINPID
+Restart=no
```
laurentfufu, line 6: in `jelastic/haproxy:2.6.1` image, there is already an override file:
```ini
[Service]
ExecStartPre=/usr/lib/systemd/scripts/haproxy_getport.sh
StartLimitBurst=500
Restart=on-failure
```

laurentfufu (changes requested):

mixins/haproxy.yml
adoux: I kept the same path that was used by Jelastic's prestart script but I changed it to `/usr/local/sbin`

packages/jahia/migrations/migrate-to-v17.yaml
```diff
@@ -96,3 +98,54 @@ actions:
         - if ("${response.out}" == "restart"):
             setGlobals:
               jahiaRollingRestartNeeded: true
+
+  upgradeHaproxy:
+  - setGlobalRepoRootUrl
+  - cmd[bl]: |-
+      REDEPLOY_FILE=/etc/jelastic/redeploy.conf
+      HAPROXY_CONF_BASE=/etc/haproxy/haproxy.cfg.jahia/jahia-cloud.cfg
+      HAPROXY_PRESTART_SCRIPT=/usr/lib/systemd/scripts/haproxy_prestart.sh
```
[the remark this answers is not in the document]
adoux: I kept the same path that was used by Jelastic's prestart script but I changed it to `/usr/local/sbin`

assets/haproxy/haproxy_service_override
adoux: Yes but it's useless, the StartLimitBurst makes the service restart 500 times if the conf is broken, which produces a lot of journald logs, and the `haproxy_getport.sh` is useless for us, it only sets 2 envvars, one being unused (the port), one being the equivalent of `haproxy_cfg_timeout_server` that we already have and that is configured on all environments so I don't think it brings anything to change it

laurentfufu: ok, thanks for the explanation :)

mixins/haproxy.yml
laurentfufu: the best practice is to put this kind of custom script in `/usr/local/s?bin`, that's why :stuck_out_tongue:

laurentfufu (approved):

PR #425, opened 2022-07-14
PAAS-2106: Update EC score/topology before restore if needed

amitc-jahia:
JIRA issue: https://jira.jahia.org/browse/PAAS-2106

Short description:

### DIFF
--- a/packages/common/restore.yml
+++ b/packages/common/restore.yml
@@ -488,8 +488,22 @@ actions:
             backupName: ${settings.backup_name}
             logsPath: ${globals.logsPath}
 
+    - vaultSecretReadKeyB64:
+        secretPath: paas/envs-common/cloud-ec-infra-circleci-api
+        secretKey: token
+    - set:
+        circleCiToken: ${globals.vaultSecretData}
+    - if ("${globals.papiToken.print()}" == ""):
+        getPapiInfoAll
+    - getCloudConf
     - cmd[${nodes.cp.first.id}]: |-
         ## [${globals.logAction}] - 4/5
+        source /metadata_from_HOST
+        if [ -z $JEL_ENV_ROLE ]; then
+          echo "JEL_ENV_ROLE variable is not defined in /metadata_from_HOST"
+          exit 1
+        fi
+
         output_file=$(mktemp)
         ec_admin_credentials=${globals.ecAdminCredentials}
         timestamp=$(echo "${settings.timestamp}"| awk '{print tolower($0)}')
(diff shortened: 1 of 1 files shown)


packages/common/restore.yml
```diff
@@ -499,9 +513,106 @@ actions:
...
+
+          pipelineId=$(echo $response | jq -r .id)
+          response=$(curl -LSs "https://circleci.com/api/v2/pipeline/$pipelineId/workflow" -H "Circle-Token: $CCToken")
+          if [ "$(echo $response | jq -r '.items[0].status')" == "null" ]; then
+            echo "No workflow created on CircleCI, aborting. Pipeline ID: $pipelineId. Response: $response" >&2
+            exit 1
+          fi
+
+          # Wait for max 10 minutes for EC deployment update to finish then restore the snapshot anyways
+          timeout=600 # 10 minutes
```
Ronan-Gorain, line 597: It takes more than 10 min for an ES cluster to resize. I suggest to set the same value as the one in runESCircleCIPipeline in mixin/elasticsearch.yaml (20 minutes)

Ronan-Gorain (changes requested):

packages/common/restore.yml
amitc-jahia: Yes it does take longer than 10 minutes. Discussed with @laurentfufu that 20 minutes wait would be long for the restore process and decided to have a wait of 5-10 minutes and then restore the backup anyways.

@laurentfufu - should we have it as 20 instead of 10?

laurentfufu: Yes, the maximum 10 minutes wait allows us to have the assurance of having at least 1 node corresponding to the new topology and the work of copying the shards (in the case of a shared network) already well advanced. The idea here is not to waste too much time when restoring a snapshot.

Ronan-Gorain (approved):

packages/common/restore.yml
```diff
@@ -499,9 +513,106 @@ actions:
...
+
+        if [ "$ec_deployment_mutualized" = true ]; then
+          eval ratio=\$${JEL_ENV_ROLE}_mutualized
+        else
+          eval ratio=\$${JEL_ENV_ROLE}_dedicated
+        fi
+
+        # Get the total shards and score before the restore
+        current_total_shards=$(curl -sS -u $ec_admin_credentials https://$UNOMI_ELASTICSEARCH_ADDRESSES/_cluster/stats | jq -r '.indices.shards.total')
+        current_score=$(( (current_total_shards/ratio) + ((current_total_shards%ratio) > 0 )))
```
laurentfufu, line 542: Is this really working ?

you should better use bash _indirect reference_ here:
```suggestion
        if $ec_deployment_mutualized; then
          ratio=${JEL_ENV_ROLE}_mutualized
        else
          ratio=$${JEL_ENV_ROLE}_dedicated
        fi

        # Get the total shards and score before the restore
        current_total_shards=$(curl -sS -u $ec_admin_credentials https://$UNOMI_ELASTICSEARCH_ADDRESSES/_cluster/stats | jq -r '.indices.shards.total')
        current_score=$(( (current_total_shards/${!ratio}) + ((current_total_shards%${!ratio}) > 0 )))
```

laurentfufu (changes requested):

packages/common/restore.yml
amitc-jahia: This actually works as confirmed.

PR #453, opened 2022-08-03
PAAS-2126: Mute env-wide monitors on jCustomer stop/delete/upgrade

kevinvandel:
JIRA issue: https://jira.jahia.org/browse/PAAS-2126

### DIFF
--- a/packages/jcustomer/migrations/migrate-to-v15.yaml
+++ b/packages/jcustomer/migrations/migrate-to-v15.yaml
@@ -1,0 +1,39 @@
+---
+type: update
+version: 1.5.2
+name: Migrate jCustomer env to v14
+id: migrate-jcustomer-env-v14
+
+# Required for healthchecks
+mixins:
+  - "../../../mixins/common.yml"
+  - "../../../mixins/jcustomer.yml"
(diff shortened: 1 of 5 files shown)


packages/jcustomer/migrations/migrate-to-v15.yaml
```diff
@@ -0,0 +1,39 @@
...
+id: migrate-jcustomer-env-v14
+
+# Required for healthchecks
+mixins:
+  - "../../../mixins/common.yml"
+  - "../../../mixins/jcustomer.yml"
+  - "../../../mixins/elasticsearch.yml"
+
+globals:
+  version: 14
```
Ronan-Gorain, line 14: Shouldn't it be 15 ?

Ronan-Gorain (changes requested):

packages/jcustomer/migrations/migrate-to-v15.yaml
kevinvandel: Yes it should, nice catch !

Ronan-Gorain (approved):

PR #467, opened 2022-08-10
PAAS-2145 fix redeploy/upgrade breaking APM, set apm in setenv before restart karaf on install

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-2145

Short description:

### DIFF
--- a/packages/jcustomer/migrations/migrate-to-v15.yaml
+++ b/packages/jcustomer/migrations/migrate-to-v15.yaml
@@ -12,16 +12,24 @@ mixins:
 
 globals:
   version: 15
+  jCustomerRollingRestartNeeded: false
 
 onInstall:
   ### Pre-checks
   - checkEnvVersion: ${globals.version}
   - checkJcustomerHealth: cp
   - checkJcustomerDatadogCustomChecks: cp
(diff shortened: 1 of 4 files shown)


kevinvandel (changes requested):
We need to add an action to the migration package to reinstall the APM when it's missing

laurentfufu:
a fixAPM action have been added to migration v15 manifest

packages/jcustomer/migrations/migrate-to-v15.yaml
```diff
@@ -37,3 +45,72 @@ actions:
...
+        }
+        """
+
+    - script: |-
+        const rep = ${response.out.toJSON()}
+        const result = {"result": 0}
+        return Object.assign({}, result, rep)
+
+    - set:
+        apm_loaded: ${response.apm_loaded}
```
Ronan-Gorain, line 92: This variables seems to be unused

laurentfufu: it is now 6 feet under

Ronan-Gorain (approved):

PR #468, opened 2022-08-11
Paas 2144 pat120

kevinvandel:
JIRA issue: https://jira.jahia.org/browse/PAAS-2144

### DIFF
--- a/packages/jahia/consistency-check.yml
+++ b/packages/jahia/consistency-check.yml
@@ -107,29 +107,29 @@ actions:
         moduleSymname: personal-api-tokens
     - if( "${globals.moduleState}" == "started" ):
         - getJahiaVersion
-        - isVersionStrictlyLower:
+        - isVersionLowerOrEqual:
             a: ${globals.jahiaVersion}
-            b: 8.0.0.0
-            res: isLowerThan8000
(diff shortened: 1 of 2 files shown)


packages/jahia/consistency-check.yml
```diff
@@ -101,25 +102,34 @@ actions:
...
     - basicModuleCheck:
         moduleSymname: personal-api-tokens
     - if( "${globals.moduleState}" == "started" ):
         - getJahiaVersion
         - isVersionStrictlyLower:
             a: ${globals.jahiaVersion}
             b: 8.0.0.0
-            res: isLowerThan8100
+            res: isLowerThan8000
         - if( ${globals.isLowerThan8100} ):
```
Ronan-Gorain, line 114: This variable is not declared yet in this action, and is defined line 116
I feel like there is a confusion with variables, the  previous line should not be updated but the b variable 2 lines before should.

Ronan-Gorain (changes requested):

packages/jahia/consistency-check.yml
kevinvandel: Fixed

Ronan-Gorain (approved):

PR #469, opened 2022-08-12
PAAS-2132: Increase jdbc maxAge setting from 5min to 15min

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-2132

Short description:

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -1117,7 +1117,7 @@ actions:
         if [[ "$my_node_name_index" == "galera_${this.target}" ]]; then
           exit 0
         fi
-        timeout=600 # wait for 600s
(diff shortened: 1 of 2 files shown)


kevinvandel (changes requested):
Shouldn't an action b added to the migration package to update existing envs ?

Ronan-Gorain:
It should be updated by events. Maybe I should mention it with a comment in the package

kevinvandel:
@Ronan-Gorain I was talking about the `maxAge` update in `context.xml` (cf. https://github.com/Jahia/jahiastic-jahia/pull/210)

Ronan-Gorain:
My bad. Done

kevinvandel (approved):

PR #470, opened 2022-08-16
Paas 2028 upgrade python

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2028

Short description:

### DIFF
--- a/mixins/common.yml
+++ b/mixins/common.yml
@@ -385,6 +385,27 @@ actions:
         [[ "${packages_to_install[@]}" == "" ]] || yum --setopt=tsflags=nodocs install -y ${packages_to_install[@]};
       user: root
 
+  installPython:
+    # Parameters :
+    #  - target: target node group or id
+    #  - python_major_version: major version of Python to install
+    #  - python_minor_version: minor version of Python to install
+    - cmd [${this.target}]: |-
+        PYTHON_MAJOR_VERSION=${this.python_major_version}
+        PYTHON_MINOR_VERSION=${this.python_minor_version}
+        curl -fLSso /root/python-${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION}.rpm "http://paas-rpms.s3.amazonaws.com/python-${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION}.rpm" || exit 1
+        rpm -ivh /root/python-${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION}.rpm || exit 1
+        cd /usr/bin
+        ln -s /usr/local/bin/python${PYTHON_MAJOR_VERSION} python${PYTHON_MAJOR_VERSION}
+        python3 -m ensurepip --upgrade || exit 1
+        cd /usr/local/bin
+        rm -f /usr/local/bin/pip /usr/local/bin/pip${PYTHON_MAJOR_VERSION}
+        ln -s pip${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION} pip
+        ln -s pip${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION} pip${PYTHON_MAJOR_VERSION}
+        pip install --upgrade pip || exit 1
+        rm -f /root/python-${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION}.rpm
+      user: root
+
   installPapiScript:
     - setGlobalRepoRootUrl
     - cmd [${this}]: |-
--- a/packages/jahia/update-events.yml
+++ b/packages/jahia/update-events.yml
@@ -140,6 +140,13 @@ onAfterRedeployContainer[bl]:
   - if ('${event.params.nodeGroup:}'):
       # Handle RedeployContainersByGroup
       - setSudoer: ${event.params.nodeGroup}
+      - installRequiredPackages:
+          target: ${event.params.nodeGroup}
+          packages: "jq"
+      - installPython:
+          target: ${event.params.nodeGroup}
+          python_major_version: 3
+          python_minor_version: 8
       - setupDatadogAgentHaproxy: ${event.params.nodeGroup}
       - environment.nodegroup.ApplyData [bl]:
           data:
@@ -148,6 +155,13 @@ onAfterRedeployContainer[bl]:
   - elif ('${event.params.nodeId:}'):
       # Handle RedeployContainerById
       - setSudoer: ${event.params.nodeId}
+      - installRequiredPackages:
+          target: ${event.params.nodeId}
+          packages: "jq"
+      - installPython:
+          target: ${event.params.nodeId}
+          python_major_version: 3
+          python_minor_version: 8
       - setupDatadogAgentHaproxy: ${event.params.nodeId}
       - environment.nodegroup.ApplyData [bl]:
           data:
@@ -161,11 +175,18 @@ onAfterScaleIn[bl]:
     user: root
 
 onAfterScaleOut[bl]:
+  - forEach(event.response.nodes):
+      - installRequiredPackages:
+          target: ${@i.id}
+          packages: "jq"
+      - installPython:
+          target: ${@i.id}
+          python_major_version: 3
+          python_minor_version: 8
   - env.control.AddContainerEnvVars[cp, proc]:
     vars: {"HAPROXY_NODES_COUNT": "${nodes.bl.length}"}
   - cmd[cp,proc]: service datadog-agent restart
     user: root
-
 
 onBeforeRedeployContainer[sqldb]:
   # disable mysql autostart because JEM will do it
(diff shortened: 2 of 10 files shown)


mixins/common.yml
```diff
@@ -385,6 +385,27 @@ actions:
         [[ "${packages_to_install[@]}" == "" ]] || yum --setopt=tsflags=nodocs install -y ${packages_to_install[@]};
       user: root
 
+  installPython:
+    # Parameters :
+    #  - target: target node group or id
+    #  - python_major_version: major version of Python to install
+    #  - python_minor_version: minor version of Python to install
+    - setGlobalRepoRootUrl
```
amitc-jahia, line 393: "setGlobalRepoRootUrl" action is not needed here

mixins/common.yml
```diff
@@ -385,6 +385,27 @@ actions:
...
+    - cmd [${this.target}]: |-
+        PYTHON_MAJOR_VERSION=${this.python_major_version}
+        PYTHON_MINOR_VERSION=${this.python_minor_version}
+        curl -fLSso /root/python-${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION}.rpm "http://paas-rpms.s3.amazonaws.com/python-${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION}.rpm" || exit 1
+        rpm -ivh /root/python-${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION}.rpm || exit 1
```
amitc-jahia, line 398: After installation, the rpm file should be removed from the node

mixins/common.yml
```diff
@@ -385,6 +385,27 @@ actions:
...
+        cd /usr/bin
+        ln -s /usr/local/bin/python${PYTHON_MAJOR_VERSION} python${PYTHON_MAJOR_VERSION}
+        python3 -m ensurepip --upgrade || exit 1
+        cd /usr/local/bin
+        rm -f /usr/local/bin/pip /usr/local/bin/pip${PYTHON_MAJOR_VERSION}
+        ln -s pip${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION} pip
+        ln -s pip${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION} pip${PYTHON_MAJOR_VERSION}
+        pip install --upgrade pip || exit 1
```
amitc-jahia, line 406: After upgrading python and pip, already installed modules are not retained

packages/jahia/update-events.yml
```diff
@@ -140,6 +140,13 @@ onAfterRedeployContainer[bl]:
   - if ('${event.params.nodeGroup:}'):
       # Handle RedeployContainersByGroup
       - setSudoer: ${event.params.nodeGroup}
+      - installRequiredPackages:
+          target: ${event.params.nodeGroup}
+          packages: "jq"
+      - installPython:
+          target: "bl, sqldb"
```
amitc-jahia, line 147: Target here should be only "bl" nodeGroup

packages/jahia/update-events.yml
```diff
@@ -148,6 +155,13 @@ onAfterRedeployContainer[bl]:
   - elif ('${event.params.nodeId:}'):
       # Handle RedeployContainerById
       - setSudoer: ${event.params.nodeId}
+      - installRequiredPackages:
+          target: ${event.params.nodeId}
+          packages: "jq"
+      - installPython:
+          target: "bl, sqldb"
```
amitc-jahia, line 162: Target here should be the nodeId

packages/jahia/update-events.yml
```diff
@@ -161,12 +175,19 @@ onAfterScaleIn[bl]:
     user: root
 
 onAfterScaleOut[bl]:
+  - forEach(event.response.nodes):
+      - installRequiredPackages:
+          target: ${@i.id}
+          packages: "jq"
+      - installPython:
+          target: "bl, sqldb"
```
amitc-jahia, line 183: Target here should be only "bl" nodeGroup

packages/jahia/update-events.yml
```diff
@@ -182,18 +203,42 @@ onAfterRedeployContainer[sqldb]:
   - if ('${event.params.nodeGroup:}'):
       # Handle RedeployContainersByGroup
       - setSudoer: ${event.params.nodeGroup}
+      - installRequiredPackages:
+          target: ${event.params.nodeGroup}
+          packages: "jq"
+      - installPython:
+          target: "bl, sqldb"
```
amitc-jahia, line 210: Target here should be only "sqldb" nodeGroup

packages/jahia/update-events.yml
```diff
@@ -182,18 +203,42 @@ onAfterRedeployContainer[sqldb]:
...
           target: ${event.params.nodeGroup}
           action: enable
   - elif ('${event.params.nodeId:}'):
       # Handle RedeployContainerById
       - setSudoer: ${event.params.nodeId}
+      - installRequiredPackages:
+          target: ${event.params.nodeId}
+          packages: "jq"
+      - installPython:
+          target: "bl, sqldb"
```
amitc-jahia, line 224: Target here should be the nodeId

packages/jahia/update-events.yml
```diff
@@ -182,18 +203,42 @@ onAfterRedeployContainer[sqldb]:
...
           target: ${event.params.nodeId}
           action: enable
 
+onAfterScaleOut[sqldb]:
+  - forEach(event.response.nodes):
+    - installRequiredPackages:
+        target: ${@i.id}
+        packages: "jq"
+    - installPython:
+        target: "bl, sqldb"
```
amitc-jahia, line 238: Target here should be only "sqldb" nodeGroup

amitc-jahia (changes requested):

mixins/common.yml
adoux: Good catch, fixed!

mixins/common.yml
adoux: Good catch, thought I did it, fixed!

mixins/common.yml
adoux: I don't think it's an issue actually.
I compared the list of modules in one of my environments running with Python 3.6 and there is 109 modules, coming from these commands:
```
python3 -m pip install requests boto3 "azure==4.0.0" adal
```
which is run by the backup/restore package so it will be installed anyway when required, and:
```
python3 -m pip install requests
```
which is run by the `installPapiScript` action that is called in `onAfterRedeployContainer` event, which means it'll be run for any container being redeployed (and I added it to the `onAfterServiceScaleOut`).

So I don't think it's an issue for redeploy/upgrades, but it was definitely missing from the migration package so I fixed it

adoux:
> The "installPapiScript" action imports the "requests" module which will not be taken into account in the new way of installing python. Retaining the already installed modules must also be addressed in cases of redeploy and upgrade(only Jahia and jCustomer nodes) too. Maybe in the "onBeforeRedeployContainer" event an action can be added to retain this information and use it in the "onAfterRedeployContainer" event.

As said in my response to one of your comment, I don't think it's an issue since the installed modules are only the ones that are installed by the backup/restore package and by the `installPapiScript` action, that are both already run during redeploy or backup/restore.
Do you agree or have I misunderstood something?

packages/jahia/update-events.yml
adoux: Stupid copy/paste, good call! Fixed

packages/jahia/update-events.yml
adoux: Stupid copy/paste, good call! Fixed

packages/jahia/update-events.yml
adoux: Stupid copy/paste, good call! Fixed

packages/jahia/update-events.yml
adoux: Stupid copy/paste, good call! Fixed

packages/jahia/update-events.yml
adoux: Stupid copy/paste, good call! Fixed

packages/jahia/update-events.yml
adoux: Stupid copy/paste, good call! Fixed

mixins/common.yml
amitc-jahia: Yes indeed, my bad i missed the generic onAfterRedeployContainer event.

amitc-jahia:
> > The "installPapiScript" action imports the "requests" module which will not be taken into account in the new way of installing python. Retaining the already installed modules must also be addressed in cases of redeploy and upgrade(only Jahia and jCustomer nodes) too. Maybe in the "onBeforeRedeployContainer" event an action can be added to retain this information and use it in the "onAfterRedeployContainer" event.
> 
> As said in my response to one of your comment, I don't think it's an issue since the installed modules are only the ones that are installed by the backup/restore package and by the `installPapiScript` action, that are both already run during redeploy or backup/restore. Do you agree or have I misunderstood something?



Yes indeed, my bad i missed the generic onAfterRedeployContainer event.

amitc-jahia (approved):

PR #475, opened 2022-08-19
Paas 2084

amitc-jahia:
JIRA issue: https://jira.jahia.org/browse/PAAS-2084

Short description:

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -556,6 +556,8 @@ actions:
     # Parameters:
     # - moduleSymname: the symbolic name of the module to upgrade (for instance "distributed-sessions")
     # - moduleVersion: the version of the module to install
+    # - moduleGroupId: group id of the artifact
+    # - moduleRepository: repository that the artifact is contained in Nexus
     - set:
         moduleInstalled: true
     - checkModule:
@@ -564,12 +566,16 @@ actions:
         - installModule:
             moduleSymname: ${this.moduleSymname}
             moduleVersion: ${this.moduleVersion}
+            moduleGroupId: ${this.moduleGroupId}
+            moduleRepository: ${this.moduleRepository}
             startModule: true
     - else:
         - if ("${globals.moduleState}" == "installed"):
             - installModule:
                 moduleSymname: ${this.moduleSymname}
                 moduleVersion: ${this.moduleVersion}
+                moduleGroupId: ${this.moduleGroupId}
+                moduleRepository: ${this.moduleRepository}
                 startModule: true
         - if ("${globals.moduleState}" == "started"):
             - isVersionStrictlyHigher:
@@ -580,6 +586,8 @@ actions:
                 - installModule:
                     moduleSymname: ${this.moduleSymname}
                     moduleVersion: ${this.moduleVersion}
+                    moduleGroupId: ${this.moduleGroupId}
+                    moduleRepository: ${this.moduleRepository}
                     startModule: true
             # If the module is already running with the desired version or a more recent one, we don't need to check the status right after
             - else:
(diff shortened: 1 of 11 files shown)


mixins/jahia.yml
```diff
@@ -991,6 +992,23 @@ actions:
...
+    - getJahiaVersion
+    - isVersionHigherOrEqual:
+        a: ${globals.jahiaVersion}
+        b: 8.0.0.0
+        res: jahia8
+    - if (${globals.jahia8}):
+        - cmd[proc]: |-
+            config_file="/data/digital-factory-data/karaf/etc/org.jahia.services.env.cfg"
+            if [ ! -f $config_file ]; then
+              organization=$(echo "${user.email}" | cut -d'@' -f1)
```
kevinvandel, line 1032: Can't we use the `ORGANIZATION_NAME` node group data instead?

packages/jahia/migrations/migrate-to-v18.yaml
```diff
@@ -123,3 +124,25 @@ actions:
...
+    - if (${globals.jahia8}):
+        - installModule:
+            moduleSymname: jahia-user-feedbacks
+            moduleVersion: 1.1.0
+            moduleGroupId: org.jahia.modules
+            moduleRepository: jahia-releases
+        - cmd[proc]: |-
+            config_file="/data/digital-factory-data/karaf/etc/org.jahia.services.env.cfg"
+            if [ ! -f $config_file ]; then
+              organization=$(echo "${user.email}" | cut -d'@' -f1)
```
kevinvandel, line 143: Same thing, I think we should use `ORGANIZATION_NAME` node group data instead.

packages/common/restore.yml
```diff
@@ -187,6 +188,17 @@ actions:
             - setGlobals:
                 restoreAS: true
 
+  saveScreebConfig:
```
kevinvandel, line 191: This was the job of https://jira.jahia.org/browse/PAAS-2105 :stuck_out_tongue:

kevinvandel (changes requested):

mixins/jahia.yml
amitc-jahia: Yes indeed fetching it from node group data is much better. Fixed

packages/jahia/migrations/migrate-to-v18.yaml
amitc-jahia: Fixed

packages/common/restore.yml
amitc-jahia: As discussed, the backup-restore part for Screeb will be handled in this ticket.

packages/jahia/migrations/migrate-to-v18.yaml
```diff
@@ -117,3 +124,36 @@ actions:
...
+        a: ${globals.jahiaVersion}
+        b: 8.0.0.0
+        res: jahia8
+    - if (${globals.jahia8}):
+        - installModule:
+            moduleSymname: jahia-user-feedbacks
+            moduleVersion: 1.1.0
+            moduleGroupId: org.jahia.modules
+            moduleRepository: jahia-releases
+        - script: |-
```
Ronan-Gorain, line 140: It could be better to reuse setJahiaUserFeedbacksConfig
Doesn't matter if the version check is done twice, code is not duplicated

mixins/jahia.yml
```diff
@@ -991,6 +1019,34 @@ actions:
...
+            return {
+              "result": 0,
+              "onAfterReturn": {
+                "set": {
+                  "organizationName": orgName
+                }
+              }
+            }
+        - cmd[proc]: |-
+            config_file="/data/digital-factory-data/karaf/etc/org.jahia.services.env.cfg"
```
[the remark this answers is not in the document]
amitc-jahia: Indeed, better to have a check before adding the config lines.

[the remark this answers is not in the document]
amitc-jahia: Fixed

packages/jahia/migrations/migrate-to-v18.yaml
amitc-jahia: Fixed

Ronan-Gorain (approved):

Ronan-Gorain (approved):

PR #487, opened 2022-08-30
PAAS-2147: install and configure kibana-dashboard-provider if upgrading from Jahia 7 to Jahia 8

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2147

Short description:

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -1709,7 +1709,35 @@ actions:
     - setGlobals:
         kibanaEndpoint: "${response.successText.fromBase64()}"
 
-  configureKibanaDashboard:
+  setupKibanaDashboardAndCreateCustomerAccount:
+  # Parameters
+  # envVersion (default: the env version): the environment version (useful for upgrades)
+  - if ("${this.envVersion.print()}" == ""):
+    - getJahiaVersion
+    - set:
+        version: ${globals.jahiaVersion}
+  - else:
+      set:
+        version: ${this.envVersion}
+  - isVersionHigherOrEqual:
+      a: ${this.version}
+      b: 8.0.0.0
+      res: jahia8
+  - if (${globals.jahia8}):
+    - getEnvLinkedJcustomer:
+        envName: ${env.envName}
+    - if (${globals.unomi_env_linked}):
+      - setupKibanaDashboard:
+          roleAndAccountName: ${env.envName}-kibana-dashboard
+          __secret__password: ${fn.password(20)}
+          jcustomerEnv: ${globals.unomi_env_name}
+      - createKibanaCustomerUserAccount:
+          accountName: ${env.envName}-kibana-user
+          rolesList: ${env.envName}-kibana-dashboard
+          __secret__password: ${fn.password(20)}
+          jcustomerEnv: ${globals.unomi_env_name}
+
+  setupKibanaDashboard:
     # Parameters
     # roleAndAccountName: kibana role and es account name
     # __secret__password: es account password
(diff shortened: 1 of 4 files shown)


mixins/jahia.yml
```diff
@@ -1709,6 +1709,42 @@ actions:
...
+        return {
+          "result": 0,
+          "onAfterReturn": {
+            set: {
+              "jCustomerLinked": jCustomerLinked
+            }
+          }
+        };
+
+        return {"result": 0, "out": jCustomerLinked};
```
kevinvandel, line 1740: Why not use `getEnvLinkedJcustomer` action ?

mixins/jahia.yml
```diff
@@ -1709,6 +1709,42 @@ actions:
     - setGlobals:
         kibanaEndpoint: "${response.successText.fromBase64()}"
 
+  setupKibanaDashboard:
```
kevinvandel, line 1712: I'm not sure about this mixin, because:
1. it is used only once
2. the name is very confusing with `configureKibanaDashboard`

In the end I think it's better to do things directly in the upgrade package.

mixins/jahia.yml
```diff
@@ -1709,6 +1709,42 @@ actions:
...
+          }
+        };
+
+        return {"result": 0, "out": jCustomerLinked};
+
+    - if ("${response.out}" != ""):
+        configureKibanaDashboard:
+          roleAndAccountName: ${env.envName}-kibana-dashboard
+          __secret__password: ${fn.password(20)}
+          jcustomer_env: ${this.jCustomerLinked}
```
kevinvandel, line 1746: Ugh, I didn't see that one before, but the case of `jcustomer_env` sucks :grimacing: (should be `jCustomerEnv`)

kevinvandel (changes requested):

mixins/jahia.yml
kevinvandel: Renaming done in this PR: https://github.com/Jahia/jelastic-packages/pull/493
You'll have to rebase and take that into account.

mixins/jahia.yml
adoux: Good call, fixed

mixins/jahia.yml
adoux: I renamed both actions so it's not confusing anymore, but I keep the action in a mixin since it's used in the upgrade package and the migrate-to-18 one

mixins/jahia.yml
adoux: Done

packages/jahia/upgrade.yml
```diff
@@ -55,6 +55,17 @@ onInstall:
   - browsingNodesBulkRedeploy:
       targetDockerTag: ${this.targetDockerTag}
 
+  # If we're upgrading from Jahia 7 to Jahia 8, then we need to install and configure kibana-dashboard-provider.
+  # The setupKibanaDashboardAndCreateCustomerAccount action will make sure that the new version is >= 8.0.0.0 and that the env is linked to a
+  # jCustomer prior to doing anything
+  - isVersionStrictlyLower:
+      a: ${nodes.proc.first.customitem.nodeVersion}
```
kevinvandel, line 62: It's very dangerous to rely on the placeholder `nodeVersion` here, we should do it differently (currently it contains the value we expect, but this is an accident and Jelastic is supposed to fix it).

kevinvandel (changes requested):

packages/jahia/upgrade.yml
adoux: Good catch, fixed

kevinvandel (approved):

PR #496, opened 2022-09-07
PAAS-2151: check if module started after installation, refactor checkModule action

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2151

Short description:

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -685,9 +685,9 @@ actions:
     #   - "started" if the module is running
     #   - "installed" if the module is installed but not running
     #   - "uninstalled" if the module is not installed
-    # - globals.mostRecentInstalledVersion: the most recent version installed
-    # - globals.runningVersion: the version running (if any)
-    # - globals.installedVersionsCount: the number of installed versions
+    # - globals.mostRecentInstalledVersion: the most recent version installed (or an empty string if the module is uninstalled)
+    # - globals.installedVersionsCount: the number of installed versions (or an empty string if the module is uninstalled)
+    # - globals.runningVersion: the version running (or an empty string if the module is not running)
     - setGlobals:  # re-initialise globals to prevent successive calls to this action from overlapping
         moduleState: ""
         runningVersion: ""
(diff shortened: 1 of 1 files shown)


mixins/jahia.yml
```diff
@@ -869,13 +853,40 @@ actions:
...
+          - checkVersionOfModule:
+              moduleSymname: ${this.moduleSymname}
+              moduleVersion: ${this.moduleVersion}
+          - cmd [proc]: 'echo "Try #${@i}"; sleep 5'
+      - if ("${globals.moduleVersionState}" != "started"):
+        - log: ${this.moduleSymname}/${this.moduleVersion} has been installed
+      - else:
+        - return:
+            type: error
+            message: "An error occurred during module startup."
```
kevinvandel, line 869: This needs to be done only when `startModule` is true.
Maybe we can factorize the code by doing the check at the end of the action, outside the "if" condition.

mixins/jahia.yml
```diff
@@ -699,57 +699,41 @@ actions:
...
-        - cmd [proc]: |-
-            __secret__pat_token="${globals.__secret__pat_token}"
-            resp=$(curl -SsH "Authorization: APIToken $__secret__pat_token" "localhost/modules/api/bundles/${this.moduleSymname}/*/_localInfo")
             module_started=$(jq -r 'to_entries[] | "\(.key);\(.value | .moduleState)"' <<< $resp | grep STARTED)
-            # If module is started, then we get the running version
-            running_version=""
             if [ $? -eq 0 ]; then
+              state=started
               running_version=$(echo $module_started | cut -d';' -f1 | awk -F'/' '{print $NF}')
+              running_version_key_value="'runningVersion': '$running_version',"
```
kevinvandel, line 713: I don't get the point of this `running_version_key_value` variable :thinking:

kevinvandel (changes requested):

mixins/jahia.yml
adoux: Good catch, done!

mixins/jahia.yml
kevinvandel: That was not really my point. What I don't get is why you put the whole JSON string, instead of just storing the version, and then use it in the "echo", like it was before actually.

adoux: I changed the way we were doing and force the declaration of all 4 globals, with empty string if needed

kevinvandel (approved):

PR #512, opened 2022-09-16
PAAS-1974: add JVM parameters to jcustomer process

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-1974

Short description:

### DIFF
--- a/packages/jcustomer/migrations/migrate-to-v17.yaml
+++ b/packages/jcustomer/migrations/migrate-to-v17.yaml
@@ -1,0 +1,55 @@
+---
+type: update
+version: 1.5.2
+name: Migrate jCustomer env to v17
+id: migrate-jcustomer-env-v17
+
+# Required for healthchecks
+mixins:
+  - "../../../mixins/common.yml"
+  - "../../../mixins/jcustomer.yml"
+  - "../../../mixins/elasticsearch.yml"
+
+globals:
+  version: 17
+  jCustomerRollingRestartNeeded: false
+
+onInstall:
+  ### Pre-checks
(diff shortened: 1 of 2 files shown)


packages/jcustomer/migrations/migrate-to-v17.yaml
```diff
@@ -0,0 +1,55 @@
...
+  eventsUpdate:
+    install:
+      jps: "${baseUrl}/../update-events.yml"
+
+  updateJVMParameters:
+    - foreach (nodes.cp):
+        - cmd[${@i.id}]: |-
+            SETENV_FILE=/opt/jcustomer/jcustomer/bin/setenv
+            sum=$(md5sum $SETENV_FILE)
+            sed -ie 's/export JAVA_OPTS.*/export JAVA_OPTS="$JAVA_OPTS -Dlog4j2.formatMsgNoLookups=true -XX:+HeapDumpOnOutOfMemoryError -XX:+PrintGCDetails -Xloggc:\/opt\/jcustomer\/jcustomer\/data\/log\/gc.log -Xms2048M -Xmx2048M"/g' $SETENV_FILE
```
laurentfufu, line 48: I may have missed something but why not use the `JAVA_MAX_MEM` variable ?
Also, is it mandatory to have Xms=Xmx here ?

Ronan-Gorain: Xmx now uses the JAVA_MAX_MEM and I set Xms to 1G because it's what jcustomer uses on a new empty environment.

laurentfufu (approved):

PR #517, opened 2022-09-22
PAAS-2137: don't raise exception in Datadog custom checks

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2137

Short description:


packages/jahia/migrations/migrate-to-v19.yaml
```diff
@@ -71,3 +72,46 @@ actions:
...
+        augmented_search
+        jahia_node_not_in_haproxy_pool
+        proxysql_backend_missing
+        proxysql_connections_discrepancies
+        strongswan_connections_status
+      )
+      for check in ${checks[@]}; do
+        script_name=${check}.py
+        script_path=${checks_dir}/${script_name}
+        tmp_path=/tmp/${script_name}
```
Ronan-Gorain, line 89: I might miss something but I don't see the gain of using a tmp path and checking md5 sum to update the script. Why not just download it directly to the right path ?

packages/jahia/migrations/migrate-to-v19.yaml
```diff
@@ -71,3 +72,46 @@ actions:
...
+        else
+          mv -f $tmp_path $script_path
+        fi
+      done
+    user: root
+  - cmd [bl]: |-
+      checks_dir=/etc/datadog-agent/checks.d
+      script_name=haproxy_one_remaining_browsing.py
+      script_path=${checks_dir}/${script_name}
+      tmp_path=/tmp/${script_name}
```
Ronan-Gorain, line 106: Same here

packages/jahia/migrations/migrate-to-v19.yaml
adoux: I did that to try and be idempotent and not update the file if not required but it doesn't really make sense in the end since I'm downloading the file anyway to compare md5 strings.
So you're right, I updated the action!

packages/jahia/migrations/migrate-to-v19.yaml
adoux: I did that to try and be idempotent and not update the file if not required but it doesn't really make sense in the end since I'm downloading the file anyway to compare md5 strings.
So you're right, I updated the action!

Ronan-Gorain (approved):

PR #519, opened 2022-09-25
PAAS-2171: Add jahia-reset-pat-token package

kevinvandel:
JIRA issue: https://jira.jahia.org/browse/PAAS-2171

### DIFF
--- a/packages/jahia/reset-pat-token.yml
+++ b/packages/jahia/reset-pat-token.yml
@@ -1,0 +1,44 @@
+---
+type: update
+version: 1.5.2
+id: jahia-reset-pat-token
+name: Jahia reset PAT token
+description:
+    Re-creates the "Jahia Cloud Token" PAT token on all Jahia nodes using the token stored in Vault.
+    BE VERY CAREFUL when using the package, as this will overwrite the token and may bring down the env in the worst case.
+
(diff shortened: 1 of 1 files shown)


packages/jahia/reset-pat-token.yml
```diff
@@ -0,0 +1,44 @@
...
+          def tokenDetails = service.getTokenDetails(\"/users/root\", \"Jahia Cloud Token\", session);
+          service.deleteToken(tokenDetails.key, session);
+          service.tokenBuilder(\"/users/root\", \"Jahia Cloud Token\", session)
+            .setToken(\"$__secret__pat_token\")
+            .setActive(true)
+            .create();
+          session.save();
+        })
+      """ > ${groovy_file_path}
+  - checkPatGroovyScriptExecution
```
Ronan-Gorain, line 44: It may be useful to set it into haproxy en vars too (and reload haproxy)

It may also be safer and more convenient to disable the check on haproxy first, then update the token, set it in haproxy nodes, check that everything is ok and finally re-enable the check on haproxy side.

Ronan-Gorain (changes requested):

Ronan-Gorain (approved):

PR #524, opened 2022-09-30
PAAS-2178: set replication and galera_cluster to false in dd conf for single nodes

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2178

Short description:

### DIFF
--- a/mixins/mariadb.yml
+++ b/mixins/mariadb.yml
@@ -72,6 +72,11 @@ actions:
     - createDataDogUser
     - log: "## Finalize Datadog agent setup on ${this}"
     - setGlobalRepoRootUrl
+    - set:
+        cluster: "false"
+    - if (nodes.sqldb.length > 1):
+        set:
+          cluster: "true"
     - cmd [${this}]: |-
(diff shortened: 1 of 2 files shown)


mixins/mariadb.yml
```diff
@@ -107,8 +112,8 @@ actions:
         echo "      - 'env:${env.envName}'" >> /etc/datadog-agent/conf.d/mysql.d/conf.yaml
         echo "      - 'role:database'" >> /etc/datadog-agent/conf.d/mysql.d/conf.yaml
         echo "    options:" >> /etc/datadog-agent/conf.d/mysql.d/conf.yaml
-        echo "       replication: true" >> /etc/datadog-agent/conf.d/mysql.d/conf.yaml
-        echo "       galera_cluster: true" >> /etc/datadog-agent/conf.d/mysql.d/conf.yaml
+        echo "       replication: ${this.cluster}" >> /etc/datadog-agent/conf.d/mysql.d/conf.yaml
```
Ronan-Gorain, line 115: This check is always useless event with galera cluster. We don't use mysql replication

Ronan-Gorain (changes requested):

mixins/mariadb.yml
adoux: Good call

adoux: fixed

adoux:
I also added updated the v19 migration package

Ronan-Gorain (approved):

PR #544, opened 2022-10-26
PAAS-2187: rename jahia cloud token

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-2187

Short description:


packages/jahia/migrations/migrate-to-v19.yaml
```diff
@@ -240,3 +241,39 @@ actions:
...
+        else
+          jahia_7_parameter=""
+        fi
+        echo """
+          org.jahia.services.content.JCRTemplate.getInstance().doExecuteWithSystemSession({ session ->
+            def service = org.jahia.osgi.BundleUtils.getOsgiService(\"org.jahia.modules.apitokens.TokenService\"$jahia_7_parameter)
+            def tokenDetails = service.verifyToken(\"$__secret__pat_token\", session)
+            if (tokenDetails) {
+               tokenDetails = service.getTokenDetails(\"/users/root\", \"Jahia Cloud Token\", session)
+               service.deleteToken(tokenDetails.key, session)
```
kevinvandel, line 267: If you delete the token, you need to disable the HAProxy check.
Isn't there a way to simply update the name of token, to avoid having to recreate it ?

kevinvandel (changes requested):

packages/jahia/migrations/migrate-to-v19.yaml
Ronan-Gorain: Indeed, I updated the token instead of re-creating it

kevinvandel (approved):

PR #547, opened 2022-10-28
Paas 1681

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-1681

Short description:

### DIFF
--- a/packages/common/restore.yml
+++ b/packages/common/restore.yml
@@ -222,6 +222,19 @@ actions:
         - return:
             type: error
             message: "An error occurred while restoring jahia data. ${response.errOut}"
+    # Content History purge job should only be enabled on the processing node
+    - cmd [cp]: |-
+        file=/opt/tomcat/conf/digital-factory-config/jahia/applicationcontext-purge-jobs.xml
+        [ -f $file ] && mv $file ${file}-disabled
+      user: tomcat
+    - cmd [proc]: |-
+        file=/opt/tomcat/conf/digital-factory-config/jahia/applicationcontext-purge-jobs.xml
+        if ! [ -f $file ]; then
+          url=${globals.repoRootUrl}/assets/jahia/applicationcontext-purge-jobs.xml-disabled
+          curl -fSsLo ${file} $url || exit 1
+        fi
+      user: tomcat
+        
     # restore screeb and kibana dashboard config
     - if (${globals.jahia8}):
         - log: "Initial Screeb and Kibana dashboard config has been restored"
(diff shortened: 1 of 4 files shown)


laurentfufu (changes requested):
Shouldn't `applicationcontext-purge-jobs.xml` be enabled on processing during an install?

I'm also thinking about backup and restore:
* If the file is present (and normally activated because it is saved on the processing) in the backup, it should be deactivated on the browsings
* If the file is not present because of a "old" backup, we can probably forgot about browsing nodes, but it should be downloaded and activated on the processing

adoux:
> Shouldn't `applicationcontext-purge-jobs.xml` be enabled on processing during an install?
> 
> I'm also thinking about backup and restore:
> 
> * If the file is present (and normally activated because it is saved on the processing) in the backup, it should be deactivated on the browsings
> * If the file is not present because of a "old" backup, we can probably forgot about browsing nodes, but it should be downloaded and activated on the processing

Changes made

packages/common/restore.yml
```diff
@@ -223,6 +223,24 @@ actions:
...
             type: error
             message: "An error occurred while restoring jahia data. ${response.errOut}"
+    # Content History purge job should only be enabled on the processing node
+    - cmd [cp]: |-
+        file=/opt/tomcat/conf/digital-factory-config/jahia/applicationcontext-purge-jobs.xml
+        if [ -f $file ]; then
+          mv $file ${file}-disabled
+        else
+          url=${globals.repoRootUrl}/packages/jahia/migrations/v20_assets/applicationcontext-purge-jobs.xml-disabled
+          curl -fSsLo ${file}-disabled $url || exit 1
```
laurentfufu, line 233: i'm not sure it will be useful to download the file on browsing

packages/common/restore.yml
```diff
@@ -223,6 +223,24 @@ actions:
...
+        fi
+      user: tomcat
+    - cmd [proc]: |-
+        file=/opt/tomcat/conf/digital-factory-config/jahia/applicationcontext-purge-jobs.xml
+        if ! [ -f $file ]; then
+          url=${globals.repoRootUrl}/packages/jahia/migrations/v20_assets/applicationcontext-purge-jobs.xml-disabled
```
laurentfufu, line 239: I guess the file should be promoted to a real asset and should be available in `${globals.repoRootUrl}/assets/jahia`

laurentfufu (changes requested):

packages/common/restore.yml
adoux: I removed this

packages/common/restore.yml
adoux: Good call, done

laurentfufu (approved):

PR #598, opened 2022-12-22
PAAS-2255: Fix thread dumps rotation and clean up

kevinvandel:
JIRA issue: https://jira.jahia.org/browse/PAAS-2255

### DIFF
--- a/packages/jahia/migrations/migrate-to-v21.yaml
+++ b/packages/jahia/migrations/migrate-to-v21.yaml
@@ -31,6 +31,9 @@ onInstall:
   - setGlobalRepoRootUrl
   ### End Pre-checks
 
+  - fixThreadDumps                            # PAAS-2255
+
+  # Actions that require a restart
   - rollbackGC                                # PAAS-2259
 
   # Restart action:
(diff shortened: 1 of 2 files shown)


packages/jahia/migrations/migrate-to-v21.yaml
```diff
@@ -58,6 +61,14 @@ actions:
   eventsUpdate:
     install:
       jps: "${baseUrl}../update-events.yml"
+  
+  fixThreadDumps:
+    - cmd [cp, proc]: |-
+        script_name=thread-dump.sh
+        curl -fSsLo /usr/local/bin/$script_name ${globals.repoRootUrl}/packages/jahia/migrations/v21_assets/$script_name || exit 1
+        # Clean up remaining folders that should have been delete
+        find /var/tmp/cloud/thread_dumps/ -type d -mtime +1 -exec rm -rf {} \;
```
laurentfufu, line 70:
```suggestion
        find /var/tmp/cloud/thread_dumps/ -type d -mtime +1 -prune -exec rm -rf {} \+
```

laurentfufu (changes requested):

packages/jahia/migrations/migrate-to-v21.yaml
kevinvandel: Done

PR #603, opened 2022-12-23
PAAS-2243: Fix issues with dnsmasq when an IPSec connection has a sla…

amitc-jahia:
…sh in the name

JIRA issue: https://jira.jahia.org/browse/PAAS-2243

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v21.yaml
+++ b/packages/jahia/migrations/migrate-to-v21.yaml
@@ -32,6 +32,7 @@ onInstall:
   ### End Pre-checks
 
   - fixThreadDumps                            # PAAS-2255
+  - updateIPSecScript                         # PAAS-2243
 
   # Actions that require a restart
   - rollbackGC                                # PAAS-2259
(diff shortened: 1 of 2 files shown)


packages/jahia/migrations/migrate-to-v21.yaml
```diff
@@ -108,3 +109,10 @@ actions:
     - if("${response.out}" == "restart"):
         - setGlobals:
             jahiaRollingRestartNeeded: true
+
+  updateIPSecScript:
+    - cmd[cp, proc]: |-
+        target=/usr/local/bin/ipsec_updown.sh
+        curl -fLSso $target ${baseUrl}/v7_assets/ipsec_updown.sh || exit 1
```
laurentfufu, line 116:
```suggestion
        curl -fLSso $target ${baseUrl}/v21_assets/ipsec_updown.sh || exit 1
```

laurentfufu (changes requested):
we should also do something to set up the missing dnsmasq conf file otherwise the problem will only be solved the next time ipsec reconnects

packages/jahia/migrations/migrate-to-v21.yaml
amitc-jahia: Fixed. As discussed, without restarting strongswan service, steps added in the migration package to create the missing dnsmasq conf file in existing envs and reload the dnsmasq service.

kevinvandel (approved):

PR #610, opened 2023-01-19
PAAS-2193: Handle s3 buckets to store logs

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-2193

Short description:

### DIFF
--- a/assets/common/rclone-logs.sh
+++ b/assets/common/rclone-logs.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+. /.jelenv
+
+nodeId=$(hostname | sed -r 's/^node([0-9]+)-.*$/\1/')
+
+case "$_PROVIDE" in
+    haproxy)
+        log_path=/var/log/haproxy/
+        include='--include="haproxy.log*.gz"'
+        ;;
+    jahia)
+        log_path=/opt/tomcat/logs/
+        include='--include="catalina.out*.gz" --include="access_log.txt*.gz"'
+        ;;
+    unomi)
+        log_path=/opt/jcustomer/jcustomer/data/log
+        include='--include="karaf.log*.gz"'
+        ;;
+    *)
+        exit 0
+        ;;
+esac
+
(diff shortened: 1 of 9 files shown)


assets/common/rclone-logs.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+. /.jelenv
+
+nodeId=$(hostname | sed 's/^[[:alpha:]]\+\([[:digit:]]\+\).*/\1/' | tr [A-Z] [a-z])
```
laurentfufu, line 5: why convert uppercase to lowercase (from an hostname) to get the node ID ?

Also I gess something like this will be more readable
```suggestion
nodeId=$(hostname | sed -r 's/^node([0-9]+)-.*$/\1/'
```

mixins/common.yml
```diff
@@ -981,3 +981,75 @@ actions:
...
+              vars={
+                "logs_s3_bucket_name": __secret__data["__secret__bucket_name"],
+                "logs_s3_bucket_region": __secret__data["s3_bucket_region"],
+                "logs_s3_aws_access_key": __secret__data["__secret__access_key"],
+                "logs_s3_aws_secret_key": __secret__data["__secret__secret_key"]
+              }
+            );
+          }
+        });
+        return {"result": 0};
```
laurentfufu, line 1055: So the credentials will be visible in `/jahia/administration/systemInfos` ? This didn't looks a good idea to me.

I would prefer them to be hardcoded in the rclone configuration file owned by the root user.

laurentfufu (changes requested):

assets/common/rclone-logs.sh
Ronan-Gorain: Fixed

mixins/common.yml
Ronan-Gorain: Creds are now hardcoded in rclone config.
Only the bucket name remain as env var as it can't be set in rclone config

laurentfufu (approved):

PR #615, opened 2023-01-23
PAAS-2265: increase proxySQL max_connections to 100 to match jdbc limit

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-2265

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v22.yaml
+++ b/packages/jahia/migrations/migrate-to-v22.yaml
@@ -84,7 +84,7 @@ actions:
         else
           echo "JDBC configuration alreay up to date"
         fi
-        sed -i 's/max_connections =.*/max_connections = 100/g' /etc/proxysql.cnf
+        sed -i 's/max_connections = 50/max_connections = 100/' /etc/proxysql.cnf
       user: root
     - cmd[proc]: |-
         ${globals.proxysql_cli} -e "update mysql_servers set max_connections=100"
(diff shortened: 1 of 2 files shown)


packages/jahia/migrations/migrate-to-v22.yaml
```diff
@@ -84,6 +84,12 @@ actions:
         else
           echo "JDBC configuration alreay up to date"
         fi
+        sed -i 's/max_connections =.*/max_connections = 100/g' /etc/proxysql.cnf
```
laurentfufu, line 87: This file is taked into account only if proxysql is started with proxysql database yet.
Also this sed will be dangerous (if the previous situation become true) because it will also increase max_connections for the datadog user.
And `g` flag is not needed :trollface:

laurentfufu (changes requested):

packages/jahia/migrations/migrate-to-v22.yaml
Ronan-Gorain: The datadog user setting is not catched by the regex as there are multiple spaces before the "="

laurentfufu: even worst for easily understand what this sed is exactly doing (from my point of view :trollface: )

about the first comment, i wanted to say "This file is taked into account only if proxysql is started **without** proxysql database yet."

laurentfufu (approved):

PR #636, opened 2023-02-09
PAAS-2270 bump to dd agent v7

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-2270

Short description:


Ronan-Gorain (changes requested):
single char variables name (like $f) are not good because we have to check their affectation to know what they might contain

Ronan-Gorain (approved):

adoux (approved):

PR #653, opened 2023-03-16
PAAS-2310: add whitelist for SAM module

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2310

### DIFF
--- a/packages/jahia/migrations/migrate-to-v23.yaml
+++ b/packages/jahia/migrations/migrate-to-v23.yaml
@@ -37,6 +37,7 @@ onInstall:
   - removeDistribution                  # PAAS-2285
   - updateIPSecScript                   # PAAS-2293
   - removeAzurePythonPackages           # PAAS-2293
+  - updateSAMWhitelist                  # PAAS-2312
 
   # Actions that require a restart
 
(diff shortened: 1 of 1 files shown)


kevinvandel (changes requested):
While we're at it, since you implemented an easy way to update the config, could you please add database-connector and elasticsearch-connector when adding AS, and jexperience-dashboards and kibana-dashboard-provider when adding jExperience?

adoux:
> While we're at it, since you implemented an easy way to update the config, could you please add database-connector and elasticsearch-connector when adding AS, and jexperience-dashboards and kibana-dashboard-provider when adding jExperience?

I tested and having the modules in the list is not an issue when they are not installed so I got rid of that extra action and added the whole list directly to the SAM config file

packages/jahia/migrations/migrate-to-v23.yaml
```diff
@@ -133,3 +134,23 @@ actions:
...
+        if ! grep -q "probes.ModuleState.whitelist" $sam_conf_file; then
+          modules=(
+            distributed-sessions
+            personal-api-tokens
+            server-availability-manager
+            graphql-dxm-provider
+            org.jahia.bundles.clustering
+            jexperience
+            jexperience-dashboards
+            kibana-dashboards
```
kevinvandel, line 150: Isn't it `kibana-dashboards-provider` instead?

adoux: Yes it is, good call! Fixed

kevinvandel (approved):

laurentfufu (approved):

PR #658, opened 2023-03-23
PAAS-2270: fix strongswan custom check after datadog-agent upgrade to v7

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2270

Short description:
The `strongswan_connections_status.py` script was not working anymore since we upgraded Datadog to v7 (we had bytes instead of a string when running the command `strongswan statusall`.

I also took advantage of this PR to install the package `gnureadline` to fix the Python3 shell (the arrows, tabulation and other keys wouldn't not behave as expected, making the usage of REPL very difficult)

### DIFF
--- a/packages/jahia/migrations/migrate-to-v23.yaml
+++ b/packages/jahia/migrations/migrate-to-v23.yaml
@@ -69,6 +69,15 @@ actions:
       jps: "${baseUrl}../update-events.yml"
 
   updateDatadogAgent:
+    - cmd [cp, proc]: |-
+        check_filename=strongswan_connections_status.py
+        check_file=/etc/datadog-agent/checks.d/$check_filename
+        if ! (echo 88972f3df25145006bf5b05c5eb63cc8 $check_file | md5sum --status -c); then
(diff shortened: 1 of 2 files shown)


packages/jahia/migrations/migrate-to-v23.yaml
```diff
@@ -69,6 +69,15 @@ actions:
...
 
   updateDatadogAgent:
+    - cmd [cp, proc]: |-
+        check_filename=strongswan_connections_status.py
+        check_file=/etc/datadog-agent/checks.d/$check_filename
+        if ! (echo 88972f3df25145006bf5b05c5eb63cc8 $check_file | md5sum --status -c); then
+          curl -fLSso $check_file ${globals.repoRootUrl}/packages/jahia/migrations/v23_assets/$check_file || exit 1
+          chown dd-agent: $check_file
+        fi
+        pip install gnureadline
```
kevinvandel, line 79: Don't we also need to add this `gnureadline` install in install package or Docker image, to have it on new envs too?

laurentfufu (approved):

PR #662, opened 2023-03-24
PAAS-2315: fix Jahia jCustomer unlink issue

julien-gagnon:
JIRA issue: https://jira.jahia.org/browse/PAAS-2315

Short description:
Add the removeKibanaDashboardAccountsAndSpace action to make sure everything kibana related is deleted when unlinking.

### DIFF
--- a/packages/jahia/unlink-jahia-jcustomer.yml
+++ b/packages/jahia/unlink-jahia-jcustomer.yml
@@ -29,8 +29,14 @@ onInstall:
   - else:
       - setGlobals:
           jahiaEnvStatus: ${settings.jahiaEnvStatus}
-  - if ("${globals.jahiaEnvStatus}" == "1"):
-      - removeAndCleanJexperience
+  - getJcustomerEnvStatus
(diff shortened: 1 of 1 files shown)


julien-gagnon:
I preferred to not call removeKibanaDashboardAccountsAndSpace in the removeAndCleanJexperience action because I was worried this could break it's usage in common/restore.yml restoreJahia action.

laurentfufu:
As this action will failed if the jcustomer env is stopped, perhaps it would be better to test the status of the jcustomer environment beforehand so that you can eventually start it before running the `removeKibanaDashboardAccountsAndSpace` action and then stop it again if necessary.

I guess we never use them but there is endpoint for that: https://docs.jelastic.com/api/7.3.2/public/#!/api/environment.Control-method-StartEnv and https://docs.jelastic.com/api/7.3.2/public/#!/api/environment.Control-method-StopEnv

laurentfufu:
indeed if the jcustomer env is stopped, it looks like the manifest fails unless the isUnomiEnvDeletion setting is true.

julien-gagnon:
@kevinvandel here is the check we talked about this morning.

kevinvandel (approved):

laurentfufu (approved):

PR #665, opened 2023-03-29
PAAS-2332: remove info from error page

julien-gagnon:
JIRA issue: https://jira.jahia.org/browse/PAAS-2332

Short description: add v24 migration package & remove info from error page

### DIFF
--- a/packages/jahia/install.yml
+++ b/packages/jahia/install.yml
@@ -12,7 +12,7 @@ description:
     needs for sustainable growth.
 
 globals:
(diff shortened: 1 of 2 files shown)


julien-gagnon:
I saw that there's usually a check before setting jahiaRollingRestartNeeded: true, but I think that the restart will be needed no matter what in this action. Let me know if you still want a check to make sure the file was modified.

Ronan-Gorain (approved):

PR #666, opened 2023-03-30
PAAS-2300 add an ip denylist configuration

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-2300

Short description:
since a production environment has a `/etc/haproxy/blacklisted_ips.txt` file used by a custom client configuration, i preferred to stick with this path and name for the file (although i would have preferred a non-racist name).
Another environment with a manually set up denylist also uses this same name. In addition, I have manually moved the configuration block within `jahia-cloud.cfg` to the same location.

### DIFF
--- a/packages/jahia/migrations/migrate-to-v24.yaml
+++ b/packages/jahia/migrations/migrate-to-v24.yaml
@@ -31,6 +31,8 @@ onInstall:
   - setGlobalRepoRootUrl
   ### End Pre-checks
 
+  - addIpDenyList                       # PAAS-2300
+
   # Actions that require a restart
   - removeInfoFromErrorPage             # PAAS-2332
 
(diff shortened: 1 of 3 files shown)


packages/jahia/migrations/migrate-to-v24.yaml
```diff
@@ -33,6 +33,7 @@ onInstall:
 
   # Actions that require a restart
   - removeInfoFromErrorPage             # PAAS-2332
+  - addIpDenyList                       # PAAS-2300
```
Ronan-Gorain, line 36: I might be wrong, but I think this section is  for jahia restart only (in that case we should maybe add "jahia" in the comment), not haproxy reload

laurentfufu: erf, I did it too fast :grin:

Ronan-Gorain (approved):

PR #670, opened 2023-04-05
Paas 2330: remove sudo permissions

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2330

The `jelastic` user is now linked to `root` so any `cmd` that doesn't have any user set will be run with `root`.

### DIFF
--- a/mixins/common.yml
+++ b/mixins/common.yml
@@ -20,6 +20,7 @@ actions:
     - log: "## Set Jelastic user as root on ${this}"
     - cmd [${this}]: sed -i 's;^jelastic:.*;jelastic:x:0:0::/root:/bin/bash;' /etc/passwd
       user: root
+    - cmd [${this}]: curl -fLSso /etc/sudoers.d/jahia-rules ${globals.repoRootUrl}/assets/common/sudoers-file || exit 1
 
   setJournaldLimit:
     - cmd[*]: |-
--- a/mixins/common.yml
+++ b/mixins/common.yml
@@ -16,26 +16,15 @@ actions:
     - else:
         log: "[INFO] global repoRootUrl is already defined with value '${globals.repoRootUrl}'"
 
-  setSudoer:
-    - log: "## Set sudoer on ${this}"
-    - cmd[${this}]: |-
-        touch /etc/sudoers.d/sudo
-        # Warning : it only works if jelastic user exists, which may be compromised
-        # if using a non-Jelastic docker image
-        USER=$(id -nu $(id -u jelastic))
-        echo "$USER ALL=NOPASSWD: ALL" >> /etc/sudoers.d/sudo
-        echo "jelastic ALL=NOPASSWD: ALL" >> /etc/sudoers.d/sudo
+  setJelasticUserAsRoot:
+    - log: "## Set Jelastic user as root on ${this}"
+    - cmd [${this}]: sed -i 's;^jelastic:.*;jelastic:x:0:0::/root:/bin/bash;' /etc/passwd
       user: root
-    - if ("${response.errOut}" != ""):
-        - return:
-            type: error
-            message: "An error occurred during sudoers configuration."
 
   setJournaldLimit:
     - cmd[*]: |-
         sed -i 's/.*SystemMaxUse=.*/SystemMaxUse=200M/g' /etc/systemd/journald.conf
         systemctl restart systemd-journald.service
-      user: root
     - if ("${response.errOut}" != ""):
         - return:
             type: error
@@ -80,14 +69,12 @@ actions:
     - cmd[${this}]: |-
         curl -fLSso /usr/local/bin/log_event.sh ${globals.repoRootUrl}/assets/common/log_events.sh || exit 1
         chmod u+x /usr/local/bin/log_event.sh
-      user: root
 
   logEvent:
     # Check if env is running (status = 1)
     if (1 == ${env.status}):
       cmd[${this.target}]: |-
         /usr/local/bin/log_event.sh "${this.title}" "${this.text}"
-      user: root
 
   manageSystemdService:
     # Parameters:
@@ -95,7 +82,6 @@ actions:
     #   - unit
     #   - action
     - cmd[${this.target}]: systemctl ${this.action} ${this.unit}
-      user: root
 
   setCloudConf:
     # Grab dev/preprod/prod infrastructure conf file
@@ -335,8 +321,8 @@ actions:
   installBackupTools:
     - setGlobalRepoRootUrl
     - cmd [${this.target}]: |-
-        sudo python3 -m pip install --upgrade pip
-        sudo python3 -m pip install --upgrade boto3
+        python3 -m pip install --upgrade pip
+        python3 -m pip install --upgrade boto3
         [ -d jelastic_backup ] || mkdir jelastic_backup
         cd jelastic_backup
         for file in backrest.py lib_aws.py revisionNode.py; do
@@ -346,7 +332,6 @@ actions:
 
   clearJelasticLogs:
     - cmd[${this.target}]: echo "" > ${this.logsPath}; chown ${this.user}:root ${this.logsPath}
-      user: root
 
   checkDatadogAgentCheck:
     # Parameters :
@@ -361,14 +346,12 @@ actions:
           echo "[ERROR] Datadog agent's '${this.checkName}' check failed" >&2
           exit 1
         fi
-      user: root
 
   installLatestDatadogAgent:
     - cmd[${this}]: |-
         source /etc/profile
         export DD_API_KEY=$DATADOGAPIKEY
         bash -c "$(curl -L https://s3.amazonaws.com/dd-agent/scripts/install_script_agent7.sh)"
-      user: root
 
   installRequiredPackages:
     # Parameters :
@@ -384,7 +367,6 @@ actions:
           fi
         done
         [[ "${packages_to_install[@]}" == "" ]] || yum --setopt=tsflags=nodocs install -y ${packages_to_install[@]};
-      user: root
 
   installPython:
     # Parameters :
@@ -409,7 +391,6 @@ actions:
         ln -s pip${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION} pip${PYTHON_MAJOR_VERSION}
         pip install --upgrade pip || exit 1
         rm -f /root/python-${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_VERSION}.rpm
-      user: root
 
   installPapiScript:
     - setGlobalRepoRootUrl
@@ -418,7 +399,6 @@ actions:
         python3 -m pip install requests
         curl -fLSso /usr/local/bin/papi.py ${globals.repoRootUrl}/assets/common/papi.py || exit 1
         chmod 755 /usr/local/bin/papi.py
-      user: root
 
   ###################
   # Vault actions   #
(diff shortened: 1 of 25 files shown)


mixins/jahia.yml
```diff
@@ -1204,7 +1197,7 @@ actions:
             organization_name="${this.organizationName}"
             env_name="${env.envName}"
             if [ ! -f $config_file ]; then
-              touch $config_file
+              sudo -u tomcat touch $config_file
```
laurentfufu, line 1200: I'm just taking this example but this is present in several places in the PR: wouldn't it be better to add the `user: tomcat` parameter to the `cmd` block?

mixins/common.yml
```diff
@@ -335,8 +320,8 @@ actions:
   installBackupTools:
```
laurentfufu, line 320: it's not a criticism and  it's open to discussion: backups and restores will now be run by the `root` user instead of the `tomcat` user or `mysql` user?

mixins/common.yml
```diff
@@ -16,26 +16,14 @@ actions:
...
-        echo "$USER ALL=NOPASSWD: ALL" >> /etc/sudoers.d/sudo
-        echo "jelastic ALL=NOPASSWD: ALL" >> /etc/sudoers.d/sudo
-      user: root
-    - if ("${response.errOut}" != ""):
-        - return:
-            type: error
-            message: "An error occurred during sudoers configuration."
+  setJelasticUserAsRoot:
+    - log: "## Set Jelastic user as root on ${this}"
+    - cmd [${this}]: sed -i 's;^jelastic:.*;jelastic:x:0:0::/root:/bin/bash;' /etc/passwd
```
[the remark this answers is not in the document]
adoux: Stupid oversight after I updated updated the action, fixed

mixins/jahia.yml
adoux: You're right it's definitely better to keep using the application user directly, I changed this everywhere

mixins/common.yml
adoux: Yes it will.
I don't think it would be an issue but maybe I forgot about something?

If we want to keep using the application users to backup/restore (which I don't mind if there is a valid reason) then I will have to change the `jelastic_backup` folder (now located in `/root` with this PR) to something like `/var/jelastic_backup` and chown the folder to the user with uid 700 (haproxy, mysql, tomcat... Which also means changing the uid/gid of `karaf` user for jcustomer nodes). Or I could just `chmod 755` but not ideal.

Anyway, if you think it is better to not use root then I will make the appropriate changes

laurentfufu: it's ok for me, I was mainly asking if it was intentional or not :stuck_out_tongue:

mixins/jahia.yml
```diff
@@ -69,17 +67,16 @@ actions:
...
-          fi
-      user: root
+        touch "/data/digital-factory-data/[persisted-configurations].dorestore"
+        source /etc/locale.conf
+        echo "JAHIA_UPGRADE=$JAHIA_UPGRADE"
+        if [ "$JAHIA_UPGRADE" == "true" ]; then
+          echo "This is an upgrade, processing's tomcat will not be restarted now"
+        else
+          echo "This is a regular redeploy, restart tomcat now"
+          sudo service tomcat start
```
laurentfufu, line 77: Isn't the user `tomcat` supposed to have no more `sudo` rights?

Either this kind of action should be performed in `root` or the `tomcat`/`mysql` users should still be given the strict rights they need (as managing their own services).

packages/common/backup.yml
```diff
@@ -409,28 +397,27 @@ actions:
...
         mkdir $BACKUP_DIR
-        chown mysql: $BACKUP_DIR
         GALERA_INFOS=""
         if [ "$(mysql -sNe  "show global status like 'wsrep_ready'"|grep ON)" != "" ]; then
           GALERA_INFOS="--galera-info"
         fi
-        sudo -u mysql ionice -n 5 mariabackup --backup --target-dir=$BACKUP_DIR $GALERA_INFOS --compress --compress-threads=2 --user mysql 2>> ${globals.logsPath} || exit 1
-        tar -H posix -C "$PWD/$BACKUP_DIR" -cf $BACKUP_NAME . 2>> ${globals.logsPath}
+        ionice -n 5 mariabackup --backup --target-dir=$BACKUP_DIR $GALERA_INFOS --compress --compress-threads=2 --user mysql 2>> ${globals.logsPath} || exit 1
+      user: mysql
```
laurentfufu, line 408: is this one done with the `mysql` on purpose ?

laurentfufu (changes requested):

laurentfufu (changes requested):

packages/common/backup.yml
adoux: The command was run with `sudo -u mysql`.
I tried to run it as root on one of my envs and I got:
```
root@node51763-adoux1-adoux ~ $ ionice -n 5 mariabackup --backup --target-dir=/tmp/test --compress --compress-threads=2 --user mysql
[00] 2023-04-10 19:12:57 Connecting to MySQL server host: localhost, user: mysql, password: not set, port: not set, socket: not set
[00] 2023-04-10 19:12:57 Failed to connect to MySQL server: Access denied for user 'mysql'@'localhost'.
```

I know this is something we could fix but I don't see the point since we were already using `mysql` to run mariabackup

packages/common/restore.yml
```diff
@@ -430,22 +425,21 @@ actions:
...
+        service tomcat stop
         __secret__ACCESS_KEY_ID="${globals.__secret__backupAccessKeyId}"
         __secret__SECRET_ACCESS_KEY="${globals.__secret__backupSecretAccessKey}"
         export RCLONE_S3_ACCESS_KEY_ID="$__secret__ACCESS_KEY_ID"
         export RCLONE_S3_SECRET_ACCESS_KEY="$__secret__SECRET_ACCESS_KEY"
 
-        cd /opt/tomcat/temp/jelastic_backup
+        cd jelastic_backup
         rclone copy --include=digital-factory-data.tar.gz --include=digital-factory-config.tar.gz --include=jahia-env-vars.gz backup:${globals.bucketName}/${globals.backupName}/ . || exit 1
         chown tomcat: digital-factory-data.tar.gz digital-factory-config.tar.gz jahia-env-vars.gz
```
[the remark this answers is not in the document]
adoux: Nope it isn't good catch, fixed!

packages/common/restore.yml
```diff
@@ -430,22 +425,21 @@ actions:
...
-      user: root
 
     - cmd [proc,cp]: |-
         cd jelastic_backup
-        sudo rm -rf /data/digital-factory-data
-        sudo chown tomcat:tomcat /data
+        rm -rf /data/digital-factory-data
         tar xf digital-factory-data.tar.gz -C / 2>>${globals.logsPath} || exit 1
+        chown tomcat: /data
         rm digital-factory-data.tar.gz
```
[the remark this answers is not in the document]
adoux: Done

packages/common/restore.yml
```diff
@@ -430,22 +425,21 @@ actions:
...
         chown tomcat: digital-factory-data.tar.gz digital-factory-config.tar.gz jahia-env-vars.gz
-      user: root
 
     - cmd [proc,cp]: |-
         cd jelastic_backup
-        sudo rm -rf /data/digital-factory-data
-        sudo chown tomcat:tomcat /data
+        rm -rf /data/digital-factory-data
         tar xf digital-factory-data.tar.gz -C / 2>>${globals.logsPath} || exit 1
+        chown tomcat: /data
```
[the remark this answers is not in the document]
adoux: forgot the `-R` parameter

packages/common/restore.yml
```diff
@@ -469,13 +463,15 @@ actions:
...
         rm -f /data/digital-factory-data/repository/.lock
+    - cmd [proc,cp]: |-
         # Add reindex flag to trigger reindex on restart
         touch /data/digital-factory-data/repository/reindex
         touch /data/digital-factory-data/safe-env-clone
         short_name=$(echo ${_ROLE}.$HOSTNAME | sed -r 's/^([a-Z]+)\.[a-Z]+([0-9]+)-.+$/\1.\2/' | tr 'A-Z' 'a-z')
         sed -e "s|^cluster.node.serverId.*|cluster.node.serverId = $short_name|" \
             -e "s|^jahia.session.jvmRoute.*|jahia.session.jvmRoute = $short_name|" \
             -i $STACK_PATH/conf/digital-factory-config/jahia/jahia.node.properties
+      user: tomcat
```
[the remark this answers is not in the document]
adoux: That's what I did before you told me that it would be better to run the cmd block as `tomcat` as opposed to use `sudo -u tomcat` :-)
But I agree so it's fixed now

packages/common/restore.yml
```diff
@@ -469,13 +463,15 @@ actions:
         chown tomcat:tomcat -R /opt/tomcat/conf
```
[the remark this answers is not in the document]
adoux: that `chown` was useless indeed, fixed

packages/common/restore.yml
```diff
@@ -775,10 +770,9 @@ actions:
         export RCLONE_S3_ACCESS_KEY_ID="$__secret__ACCESS_KEY_ID"
         export RCLONE_S3_SECRET_ACCESS_KEY="$__secret__SECRET_ACCESS_KEY"
 
-        cd /var/lib/haproxy/jelastic_backup
+        cd jelastic_backup
         rclone copy backup:${globals.bucketName}/${globals.backupName}/haproxy.tar.gz . || exit 1
         chown haproxy: haproxy.tar.gz
```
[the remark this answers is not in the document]
adoux: Nope, good catch again

packages/common/restore.yml
```diff
@@ -819,11 +813,12 @@ actions:
           sed -i "/use_backend proc.*/c \ \ \ \ $cookie_line1" $HAPROXY_CONF
           sed -i "/cookie JSESSIONID.*/c \ \ \ \ $cookie_line2" $HAPROXY_CONF
         fi
+        chown -R haproxy: /etc/haproxy
```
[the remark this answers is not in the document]
adoux: Indeed, useless `chown`, fixed

packages/common/backup.yml
adoux: But if you think we should then I can fix this

mixins/jahia.yml
laurentfufu: sure, but I'm afraid that the `TOMCAT_SERVICE` command group that we're interested in in this example is not set up anywhere in the PR, unless it's set up automatically by jelastic (it would seem dangerous for jelastic to decide for us the system rights of the application user, but why not) ?

packages/common/backup.yml
laurentfufu: it's ok for me, I just wanted to know if it was done on purpose

packages/common/restore.yml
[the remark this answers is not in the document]
laurentfufu: still have to be sure that, with this PR, the `tomcat` user will be able to `sudo touch` :trollface:

laurentfufu (dismissed):

mixins/jahia.yml
adoux: Very good catch, fixed!

mixins/haproxy.yml
```diff
@@ -6,8 +6,15 @@ globals:
   __secret__haproxy_admin_stats_pass: ${fn.password(20)}
 
 actions:
+  setHaproxySudoPermissions:
+    - cmd [${this}]: |-
+        if ! grep -qE "^haproxy "; then
```
laurentfufu, line 11: missing the file to grep

laurentfufu (commented):
now just need a more global `set<blabla>SudoPermissions` action :)

adoux:
> now just need a more global `set<blabla>SudoPermissions` action :)

Done!

julien-gagnon (approved):

PR #676, opened 2023-04-12
PAAS-2330: fix jCustomer scale out failing because of new bl nodes

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2330

### DIFF
--- a/mixins/common.yml
+++ b/mixins/common.yml
@@ -71,6 +71,7 @@ actions:
     - cmd[${this}]: |-
         curl -fLSso /usr/local/bin/log_event.sh ${globals.repoRootUrl}/assets/common/log_events.sh || exit 1
         chmod u+x /usr/local/bin/log_event.sh
+      user: root
 
   logEvent:
     # Check if env is running (status = 1)


packages/jcustomer/update-events.yml
```diff
@@ -176,9 +180,13 @@ onBeforeScaleIn:
...
+      - script: |
+          var resp = api.env.control.GetNodeInfo("${env.envName}", session, ${@i.id})
+          return {"result": 0, "nodegroup": resp.node.nodeGroup}
+      - if ("${response.nodegroup}" == "cp"):
+          - getLogEventScript: ${@i.id}
+          - logEvent:
+              target: ${@i.id}
+              title: "Scaled out ${event.params.nodeGroup} node group"
+              text: "Node ${@i.id} has been added"
+          - installPapiScript: ${@i.id}
```
laurentfufu, line 192: can it be better to move these actions in `onAfterServiceScaleOut[cp]` ?

laurentfufu (changes requested):

laurentfufu (approved):

PR #682, opened 2023-04-17
PAAS-2219: update tomcat & openjdk

julien-gagnon:
JIRA issue: https://jira.jahia.org/browse/PAAS-2219

Short description: Update tomcat & openjdk version to 9.0.71 & 11.0.18 respectively

### DIFF
--- a/packages/jahia/migrations/migrate-to-v25.yaml
+++ b/packages/jahia/migrations/migrate-to-v25.yaml
@@ -32,14 +32,21 @@ onInstall:
   ### End Pre-checks
 
   - addBlockMaliciousXForwardedFor           # PAAS-2335
-  - upgradeSam               # PAAS-2318
+  - upgradeSam                               # PAAS-2318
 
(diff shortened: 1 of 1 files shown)


packages/jahia/migrations/migrate-to-v25.yaml
```diff
@@ -90,3 +97,13 @@ actions:
                 moduleRepository: jahia-releases
             - checkJahiaHealth: "cp, proc"
             - enableHaproxyHealthcheck
+
+  upgradeTomcat:
+    - cmd[proc]:  |-
```
Ronan-Gorain, line 102: It may be "cleaner" to rely on the TOMCAT_VERSION env var (fetched via api instead of cmd block) of jahia nodes instead of grepping on a release notes  file

Ronan-Gorain (changes requested):

Ronan-Gorain (approved):

PR #696, opened 2023-04-25
PAAS-2291: split nodes in AZs

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-2291

Short description:

### DIFF
--- a/packages/jahia/install.yml
+++ b/packages/jahia/install.yml
@@ -46,6 +46,14 @@ onBeforeInstall: |
   var __secret__proxysql_admin_password = "${globals.__secret__proxysql_admin_password}";
   var __secret__proxysql_monitoring_password = "${globals.__secret__proxysql_monitoring_password}";
   var __secret__proxysql_cluster_password = "${globals.__secret__proxysql_cluster_password}";
+  nodes_distribution = {
+      "mode": "SOFT",
+      "zones": [{
+          "az1": "zoneA",
+          "az2": "zoneB",
+          "az3": "zoneC"
+      }]
+  }
 
   // HAProxy nodes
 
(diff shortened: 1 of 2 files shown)


packages/jahia/migrations/migrate-to-v26.yaml
```diff
@@ -111,6 +112,33 @@ actions:
...
+        );
+        if (resp.result != 0)
+          errors.push(resp);
+      });
+      if (errors.length > 0)
+        return {'result': 1, 'errors': errors};
+      else
+        return {'result': 0};
+
+
```
laurentfufu, line 140:
```suggestion
    - environment.nodegroup.ApplyData [cp, bl, sqldb]:
        data:
          distribution:
            mode: SOFT
            zones:
              - az1: zoneA
                az2: zoneB
                az3: zoneC
```

packages/jahia/install.yml
```diff
@@ -66,7 +66,15 @@ onBeforeInstall: |
...
-      ]
+      ],
+    "distribution": {
+        "mode": "SOFT",
+        "zones": [{
+            "az1": "zoneA",
+            "az2": "zoneB",
+            "az3": "zoneC"
+        }]
+      }
```
laurentfufu, line 77: I think using a variable to avoid repetition here would be better

Ronan-Gorain: done

packages/jahia/migrations/migrate-to-v26.yaml
Ronan-Gorain: done

laurentfufu (approved):

PR #722, opened 2023-06-05
PAAS-2393: prevent java autorestart in jcustomer

julien-gagnon:
JIRA issue: https://jira.jahia.org/browse/PAAS-2393

### DIFF
--- a/packages/jcustomer/install.yml
+++ b/packages/jcustomer/install.yml
@@ -13,7 +13,7 @@ description:
     privacy rules (such as GDPR).
 
 globals:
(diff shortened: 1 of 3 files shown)


Ronan-Gorain (changes requested):
I think this should be added to the migrate-to-v23.yaml. v22 is currently being released

julien-gagnon:
> I think this should be added to the migrate-to-v23.yaml package. v22 is already released

Absolutely. This has been corrected.

Ronan-Gorain (approved):

PR #729, opened 2023-06-16
PAAS-2432: adjust jExperience installation

julien-gagnon:
JIRA issue: https://jira.jahia.org/browse/PAAS-2432

Short description: update jExperience installation actions to take into account jCustomer 2.0. Also update jExperience Dashboard installation actions accordingly.

### DIFF
--- a/mixins/jcustomer.yml
+++ b/mixins/jcustomer.yml
@@ -16,6 +16,7 @@ actions:
             break
           }
         }
+        var version = resp['nodes'][0]['customitem']['nodeVersion'];
         var cmd = "find -H /opt/ -mindepth 4  -type f -name setenv -exec awk -F'\"' '$1~/UNOMI_ROOT_PASSWORD_B64=$/ {print $2}' {} \\;";
         var command = toJSON([{"command": cmd}]);
         var res = jelastic.env.control.ExecCmdById(unomi_env_name, session, unomi_nodeid, command)
@@ -23,10 +24,10 @@ actions:
 
 
         const container_env_vars = jelastic.env.control.GetContainerEnvVars(
-                                     unomi_env_name,
-                                     session,
-                                     unomi_nodeid
-                                   );
+                                    unomi_env_name,
+                                    session,
+                                    unomi_nodeid
+                                  );
         const allowed_ips = container_env_vars.object.UNOMI_THIRDPARTY_PROVIDER1_IPADDRESSES;
 
         return {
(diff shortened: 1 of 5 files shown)


mixins/jcustomer.yml
```diff
@@ -329,3 +328,16 @@ actions:
     - setupDatadogAgentUnomi:
         target: ${this}
         __secret__unomi_root_password_b64: ${globals.__secret__unomi_root_password_b64}
+
+  getJcustomer:
```
kevinvandel, line 332: We don't already have a mixin for this? Btw, I think it should be in jahia mixin instead of jcustomer, since it applies to a Jahia env.

packages/jahia/upgrade.yml
```diff
@@ -59,7 +60,11 @@ onInstall:
 
   # Update jExperience if needed
   - if ("${globals.jexperienceStatus}" == "started"):
-    - getJexperienceVersion
+    - getJcustomer
+    - getUnomiDnsAndPwd:
+        unomi_env_name: ${globals.jCustomerEnv}
```
kevinvandel, line 65: What don't you do this inside _getJexperienceVersion_ method?

mixins/jcustomer.yml
julien-gagnon: We do have a mixin for this. It was an action that I moved from the package to a mixin, but I should have checked if something similar already exist.

packages/jahia/upgrade.yml
julien-gagnon: Thanks for the input. I don't know why I wanted to keep everything separate, but I think the new version is better. Let me know what you think :)

mixins/jcustomer.yml
```diff
@@ -35,14 +36,12 @@ actions:
...
-              "__secret__unomi_pwd_b64": unomi_pwd_b64
+              "__secret__unomi_pwd_b64": unomi_pwd_b64,
+              "unomi_version": version
             }
           }
         };
 
-    - setGlobals:
-        unomi_version: ${nodes.cp.first.customitem.nodeVersion}
-
```
julien-gagnon, line 45: I removed that because I did not understand it. This action is in the jCustomer mixin, but only get played on Jahia env, so the value returned here is the Jahia version. Maybe I'm not understanding something, so feel free to correct me if needed :)

kevinvandel (approved):

Ronan-Gorain (approved):

PR #733, opened 2023-06-22
PAAS-2370: handle JCR index at backup/restore to make it faster

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-2370

Short description:

### DIFF
--- a/packages/common/backup.yml
+++ b/packages/common/backup.yml
@@ -14,6 +14,7 @@ globals:
 
 mixins:
   - ../../mixins/common.yml
+  - ../../mixins/jahia.yml
   - ../../mixins/elasticsearch.yml
   - ../../mixins/mariadb.yml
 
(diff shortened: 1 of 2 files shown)


packages/common/backup.yml
```diff
@@ -111,6 +112,23 @@ actions:
...
+    - cmd[proc]: |-
+        python3 -c 'with open("/data/digital-factory-data/repository/revisionNode", "wb") as file: file.write((${response.out}).to_bytes(8, byteorder="big", signed=False))'
+        touch /data/digital-factory-data/backup-with-JCR-index
+
+        cd jelastic_backup
+        rm -f digital-factory-data.tar.gz # In case file already exists
+        tar -cz -H posix -f digital-factory-data.tar.gz -C / data/digital-factory-data/ 2>>${globals.logsPath}
+        if [ "$?" -gt 1 ]; then
+          echo "tar command return an error code greater than 1. Exiting" >>${globals.logsPath}
+          exit 1
```
kevinvandel, line 128: Isn't there a way to disable FRO in that case? Or more globally to disable it in any error case? (otherwise the env will be blocked in FRO until someone disables it)

Ronan-Gorain: Indeed, I missed this part. Fixed

laurentfufu (approved):

PR #760, opened 2023-07-26
PAAS-2390: save provided list of custom conf files upon redeploy or backup/restore

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2390

### DIFF
--- a/packages/common/restore.yml
+++ b/packages/common/restore.yml
@@ -426,40 +426,61 @@ actions:
 
   restoreJahia:
     - cmd [proc,cp]: |-
-        service tomcat stop
         __secret__ACCESS_KEY_ID="${globals.__secret__backupAccessKeyId}"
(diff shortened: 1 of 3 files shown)


packages/common/restore.yml
```diff
@@ -426,40 +426,61 @@ actions:
...
+        tarfile=jahia-custom-conf-files.tar.gz
+        cd jelastic_backup
+        if [ -f $tarfile ]; then
+          tar -xPf $tarfile -C / 2>>${globals.logsPath} || { echo "ERROR" >&2; exit 0; }
+          rm -f $tarfile
+        fi
+    - if ("${response.errOut}" != ""):
+        - return:
+            type: error
+            message: "An error occurred while restoring Jahia (data)"
```
Ronan-Gorain, line 466: Wrong message, it's not jahia data that are restored here

Ronan-Gorain (changes requested):

Ronan-Gorain (approved):

PR #776, opened 2023-08-16
PAAS-2549: complete datadog storage setup

julien-gagnon:
JIRA issue: https://jira.jahia.org/browse/PAAS-2549

Short description: 
- Modify _ROLE var for storage
- Add glusterFS log consumption configuration on client nodes

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -257,9 +257,19 @@ actions:
           sed "s/__PASSWORD__/${DB_USER_DATADOG}/" -i /etc/datadog-agent/conf.d/jahia_local_revisions_discrepancies.yaml
         fi
         /usr/local/bin/set_dd_tags.sh
-        systemctl restart crond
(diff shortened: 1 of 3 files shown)


mixins/jahia.yml
```diff
@@ -257,6 +257,15 @@ actions:
           sed "s/__PASSWORD__/${DB_USER_DATADOG}/" -i /etc/datadog-agent/conf.d/jahia_local_revisions_discrepancies.yaml
         fi
         /usr/local/bin/set_dd_tags.sh
+        setfacl -m u:dd-agent:rx /var/log/glusterfs/*.log
```
Ronan-Gorain, line 260: Shouldn't this block only concern jahia nodes using glusterfs cluster (not singleNFS) ?

julien-gagnon: Changed

Ronan-Gorain (approved):

PR #782, opened 2023-08-25
PAAS-2258: add modules check

julien-gagnon:
JIRA issue: https://jira.jahia.org/browse/PAAS-2258

Short description: Add new modules check for upgrades, migration and rolling restart/redeploy.

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -2025,3 +2025,95 @@ actions:
             setfacl -m u:dd-agent:rx /var/log/glusterfs/*.log
             sed -i '/killall.*glusterd/a \  touch \/var\/log\/glusterfs\/glusterd.log && /usr\/bin\/setfacl -m u:dd-agent:rx \/var\/log\/glusterfs\/*.log' /etc/logrotate.d/glusterfs
             systemctl restart rsyslog crond datadog-agent
+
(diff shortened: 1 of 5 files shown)


mixins/jahia.yml
```diff
@@ -2025,3 +2025,93 @@ actions:
             setfacl -m u:dd-agent:rx /var/log/glusterfs/*.log
             sed -i '/killall.*glusterd/a \  touch \/var\/log\/glusterfs\/glusterd.log && /usr\/bin\/setfacl -m u:dd-agent:rx \/var\/log\/glusterfs\/*.log' /etc/logrotate.d/glusterfs
             systemctl restart rsyslog crond datadog-agent
+
+  dumpModules:
+    # Parameters:
+    # - operation: type of operation performed
```
Ronan-Gorain, line 2031: Missing doc about checkVersion parameter

Ronan-Gorain (changes requested):

mixins/jahia.yml
julien-gagnon: Doc added

Ronan-Gorain (dismissed):

Ronan-Gorain (approved):

PR #788, opened 2023-08-30
PAAS-2433: add missing ignoreModules setting to v32 migration package

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2433


kevinvandel (changes requested):
I think there was a misunderstanding when implementing the modules check. The list of modules to ignore must not be a setting of the package. It should better be a global, because we'll have to define "statically" in each migration package.

kevinvandel (approved):

Ronan-Gorain (approved):

PR #820, opened 2023-10-09
PAAS-2633: Fix redeploy and rclone config on jCustomer nodes + add se…

kevinvandel:
…tenv file to redeploy config

JIRA issue: https://jira.jahia.org/browse/PAAS-2633


Ronan-Gorain (changes requested):
Setting setenv file path in redeploy.conf is not working because of the symlink in the path

Ronan-Gorain (approved):

PR #826, opened 2023-10-13
PAAS-2605: handle jCustomer restart gracefully

julien-gagnon:
JIRA issue: https://jira.jahia.org/browse/PAAS-2605

### DIFF
--- a/mixins/jcustomer.yml
+++ b/mixins/jcustomer.yml
@@ -257,6 +257,7 @@ actions:
   refreshUnomiAllowedIPs:
     # Parameters:
     #   - unomiEnvName (optional, defaults to current envName)
+    - setGlobalRepoRootUrl
     - if ('${this.unomiEnvName.print()}' == ''):
         set:
           unomiEnvName: ${env.envName}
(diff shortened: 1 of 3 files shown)


mixins/jcustomer.yml
```diff
@@ -305,9 +305,13 @@ actions:
           envName: ${this.unomiEnvName}
           nodeGroup: cp
           vars: {"UNOMI_THIRDPARTY_PROVIDER1_IPADDRESSES": "${this.newIPList}"}
-        - api: env.control.RestartNodes
-          envName: ${this.unomiEnvName}
-          nodeGroup: cp
+        - script: |-
+            return api.marketplace.jps.Install(
+              {
+                jps: "${baseUrl}../jcustomer/jcustomer-rolling-restart.yml",
```
Ronan-Gorain, line 311: Better to use ${globals.repoRootUrl} here. If we use this action in a migration package for example, it won't work

Ronan-Gorain (changes requested):

mixins/jcustomer.yml
julien-gagnon: Fixed

Ronan-Gorain (approved):

Ronan-Gorain (approved):

PR #846, opened 2023-11-07
PAAS-2669: try to remove AS configuration file before installing AS m…

Ronan-Gorain:
…odules in case an existing configuration was present whereas it shouldn't

JIRA issue: https://jira.jahia.org/browse/PAAS-2669

Short description:

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -1669,6 +1669,7 @@ actions:
 
   removeDefaultESConnection:
     # Delete the jahia-cloud_augmented-search connexion from elasticsearch-connector module
+    - getPatTokenAndKey
     - cmd[proc]: |-
         __secret__API_TOKEN="${globals.__secret__pat_token}"
 
@@ -1741,6 +1742,13 @@ actions:
             echo "Failed to use the $es_connection_name connection for AS: $response"
             exit 1
           fi
+
+  removeAndCleanAugmentedSearch:
+    - removeDefaultESConnection
+    - uninstallModule:
+        moduleSymname: augmented-search
+    - cmd[proc, cp]: |-
+        rm -f /data/digital-factory-data/karaf/etc/org.jahia.modules.augmentedsearch.cfg
 
   #######################
   # jexperience related #
--- a/packages/jahia/augmented-search-install.yml
+++ b/packages/jahia/augmented-search-install.yml
@@ -62,6 +62,8 @@ actions:
                 jcontent_ver: 2.7.0
 
   installModules:
+    - cmd[proc, cp]: |-
+        rm -f /data/digital-factory-data/karaf/etc/org.jahia.modules.augmentedsearch.cfg
     - if(globals.jahia810):  # if jahia 8.1.0 then graphql-dxm-provider and jahia-administration must be upgraded
         - disableHaproxyHealthcheck  # upgrading graphql temporarily break the healthcheck
         - installOrUpgradeModule:
--- a/packages/jahia/augmented-search-uninstall.yml
+++ b/packages/jahia/augmented-search-uninstall.yml
@@ -14,51 +14,12 @@ mixins:
 
 
 onInstall:
-  - deleteESConnection
-  - deleteAugSearch
+  - removeAndCleanAugmentedSearch
   - removeNodeGroupData
   - destroyESDeployment
   - disableDatadogAugSearchCheck
 
 actions:
-  deleteESConnection:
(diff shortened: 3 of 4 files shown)


kevinvandel (changes requested):
We can keep this change, but to me the root cause of the issue lies here: https://github.com/Jahia/jelastic-packages/blob/c572356b2eee386fa3efcd1532a5f05246f863bf/packages/common/restore.yml#L633-L637

This is were the config isn't removed in the first place.

mixins/jahia.yml
```diff
@@ -1742,6 +1742,13 @@ actions:
             exit 1
           fi
 
+  removedAndCleanAugmentedSearch:
```
kevinvandel, line 1745:
```suggestion
  removeAndCleanAugmentedSearch:
```

packages/jahia/augmented-search-uninstall.yml
```diff
@@ -15,7 +15,7 @@ mixins:
 
 onInstall:
   - deleteESConnection
```
kevinvandel, line 17: Is this `deleteESConnection` still required?

mixins/jahia.yml
```diff
@@ -1742,6 +1742,13 @@ actions:
...
+    - removeDefaultESConnection
```
kevinvandel, line 1746: To be safe, I think we should edit `removeDefaultESConnection` to add a call to `getPatTokenAndKey` inside.

kevinvandel (changes requested):

mixins/jahia.yml
Ronan-Gorain: Fixed

mixins/jahia.yml
Ronan-Gorain: Fixed

packages/jahia/augmented-search-uninstall.yml
Ronan-Gorain: Indeed. I misread what it was doing and I thought we should keep it. But I was wrong.
Action removed

mixins/jahia.yml
kevinvandel: It seems you didn't rename the usages

Ronan-Gorain: Fixed

packages/jahia/augmented-search-install.yml
```diff
@@ -62,6 +62,8 @@ actions:
                 jcontent_ver: 2.7.0
 
   installModules:
+    - cmd[proc, cp]: |-
+        rm -f /data/digital-factory-data/karaf/etc/org.jahia.modules.augmentedsearch.cfg
```
laurentfufu, line 66: the `cmd` will not exit 0 if the file doesn't exit

mixins/jahia.yml
```diff
@@ -1742,6 +1743,13 @@ actions:
             exit 1
           fi
 
+  removeAndCleanAugmentedSearch:
+    - removeDefaultESConnection
+    - uninstallModule:
+        moduleSymname: augmented-search
+    - cmd[proc, cp]: |-
+        rm -f /data/digital-factory-data/karaf/etc/org.jahia.modules.augmentedsearch.cfg
```
laurentfufu, line 1751: I'm not sure, but isn't it better to empty the file and then delete it to prevent jahia from recreating the file with the settings information it has in the database?

packages/jahia/augmented-search-install.yml
Ronan-Gorain: It should with the -f

mixins/jahia.yml
Ronan-Gorain: Something like empty file, wait X seconds, and then remove file ?

kevinvandel: From what I understood from previous similar questions (see https://jira.jahia.org/browse/PAAS-1618), deleting the file is safe as long as Jahia is running during the operation. Emptying the file is necessary in the case Jahia is stopped, so that at next start-up it sees the file and updates its configuration in DB.
Since `removeAndCleanAugmentedSearch` is always called when Jahia is started, I think we're good.

packages/jahia/augmented-search-install.yml
laurentfufu: indeed :upside_down_face:

kevinvandel (approved):

laurentfufu (approved):

PR #887, opened 2023-12-05
PAAS-2723: no Elastic Cloud related actions if customer is LOIM

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2723

### DIFF
--- a/packages/common/restore.yml
+++ b/packages/common/restore.yml
@@ -745,61 +745,63 @@ actions:
           ec_deployment_name=$(echo $ec_deployment | jq -r .name)
           papi.py -X PUT "ec-deployment/$ec_deployment_id" -d '{"score":'$new_score'}'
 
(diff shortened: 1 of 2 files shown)


packages/common/restore.yml
```diff
@@ -745,61 +745,63 @@ actions:
           ec_deployment_name=$(echo $ec_deployment | jq -r .name)
           papi.py -X PUT "ec-deployment/$ec_deployment_id" -d '{"score":'$new_score'}'
 
-          __secret__CCToken=$(echo -n ${this.__secret__circleCiToken} | base64 -d)
+          if ("${globals.organizationName}" != "loim"):
```
Ronan-Gorain, line 748: Not sure it's bash :D

Ronan-Gorain (changes requested):

packages/common/restore.yml
laurentfufu: pretty sure it's not :trollface:

PR #892, opened 2023-12-12
PAAS-2691: fix Datadog product tag

julien-gagnon:
JIRA issue: https://jira.jahia.org/browse/PAAS-2691

### DIFF
--- a/packages/jahia/install.yml
+++ b/packages/jahia/install.yml
@@ -12,7 +12,7 @@ description:
     needs for sustainable growth.
 
 globals:
-  jahia_env_version: 37
+  jahia_env_version: 38
   db_user: jahia-db-${fn.random}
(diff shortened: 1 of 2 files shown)


Ronan-Gorain (changes requested):
The /etc/datadog-agent/datadog.yaml file is generated by a script located in /usr/local/bin/set_dd_tags.sh. I think the fix should be done in this script instead of the datadog.yaml file. Otherwise, I fear it will be wrong again the minute after

From what I see, the line 
`echo " - product:${_PROVIDE}"`
should be replaced by 
`echo " - product:$(awk '$2~/^product/ {split($2,a,":"); print a[2]}' $ddconffile)"`

Ronan-Gorain (approved):

kevinvandel (approved):

PR #900, opened 2023-12-19
PAAS-2739 retry sending logs to logs bucket if an error occurs

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-2739

Short description:

### DIFF
--- a/assets/common/rclone-logs.sh
+++ b/assets/common/rclone-logs.sh
@@ -22,4 +22,9 @@ case "$_PROVIDE" in
         ;;
 esac
 
-eval rclone copy $log_path $include logsbucket:$logs_s3_bucket_name/$envName/$_ROLE/$nodeId/
+sendLogs(){
+    eval rclone copy $log_path $include logsbucket:$logs_s3_bucket_name/$envName/$_ROLE/$nodeId/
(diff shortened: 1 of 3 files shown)


kevinvandel (approved):

packages/jcustomer/migrations/migrate-to-v27.yaml
```diff
@@ -93,6 +94,15 @@ actions:
         sed '/logs_enabled:/i echo "site: ${DD_SITE}"' -i $tagFile
       fi
 
+  updateSendLogsCron:
+    cmd[bl, cp, proc, sqldb, storage]: |-
```
Ronan-Gorain, line 98: There are no such node groups in jcustomer environments :)

Ronan-Gorain (changes requested):

packages/jcustomer/migrations/migrate-to-v27.yaml
laurentfufu: indeed !

also move to migration v28 manifest :grin:

Ronan-Gorain (approved):

PR #911, opened 2024-01-16
PAAS-2657: add new check to count indices size per jcustomer env

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2657

### DIFF
--- a/packages/jcustomer/migrations/v28_assets/jcustomer_custom_metrics.py
+++ b/packages/jcustomer/migrations/v28_assets/jcustomer_custom_metrics.py
@@ -23,6 +23,7 @@ class CheckCustomMetrics(AgentCheck):
 
     def check(self, instance):
         self.count_event(instance)
+        self.disk_usage()
 
     def count_event(self, instance):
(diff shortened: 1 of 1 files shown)


packages/jcustomer/migrations/v28_assets/jcustomer_custom_metrics.py
```diff
@@ -40,5 +41,28 @@ def count_event(self, instance):
...
+            response = requests.get(url, auth=(self.es_username, self.es_password), params=params)
+            indices = response.json()
+        except ValueError as exception:
+            self.__set_error("jcustomer.event_count: Returned invalid json")
+            return
+        except RequestException as exception:
+            self.__set_error("jcustomer.event_count: Request failed")
+            return
+        except Exception as exception:
+            self.__set_error("jcustomer.event_count: An unexpected error occured")
```
julien-gagnon, line 57: I wonder if we should set the errors to jcustomer.disk_space here.

julien-gagnon (approved):

PR #915, opened 2024-01-24
PAAS-2688: fix AS install because of esEndpoint variable

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2688

### DIFF
--- a/packages/jahia/augmented-search-install.yml
+++ b/packages/jahia/augmented-search-install.yml
@@ -125,7 +125,7 @@ actions:
         {
           "id": "${globals.esConnectionName}",
           "isConnected": true,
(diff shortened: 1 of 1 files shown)


packages/jahia/augmented-search-install.yml
```diff
@@ -125,7 +125,7 @@ actions:
         {
           "id": "${globals.esConnectionName}",
           "isConnected": true,
-          "host": "$es_endpoint",
+          "host": "${this.es_endpoint}",
```
laurentfufu, line 128:
```suggestion
          "host": "${this.esEndpoint}",
```

laurentfufu (approved):

PR #921, opened 2024-01-30
PAAS-2658: add new checks for Jahia env disk space

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2658

Short description:
* I decided to add the AS indices check to the existing AS custom check to make things easier, and the datastore check needs to be run on a storage node
    * Since checks are not running on the same nodes, we can't have a single metric so I decided to a metric that will look like `jahia.custom_metrics.customer_disk_usage.<datastore|database|augmented_search>` and we will have to add the numbers in the widget.
* For the datastore check on storage nodes, I decided to give `dd-agent` the permission to run `sudo du`, the other option was to run the `du` command as a cron (run by root) and put the result in a file that can then be read by dd-agent. **If you prefer this solution** let me know and I will do the required changes
* I decided to go for Bytes and not Megabytes for all 3 metrics to avoid potential number/unit issues
* Last but not least, since the `jahia_custom_metrics` check runs every hour (as requested in PAAS-2677) so it means the database check will run every hour as well. I did the same thing for the datastore check but the AS indices check runs every 15s so this one will run much more often
    * Let me know if you want to change this

### DIFF
--- a/packages/jahia/migrations/migrate-to-v38.yaml
+++ b/packages/jahia/migrations/migrate-to-v38.yaml
@@ -36,7 +36,7 @@ onInstall:
   - updateSendLogsCron                               # PAAS-2739
   - getLogEventScript: bl, proc, cp, sqldb, storage  # PAAS-2746 (action from mixins/common.yml)
   - updateCurl                                       # PAAS-2746
-  - addConsumptionMetricsCheck                       # PAAS-2678, PAAS-2688
+  - addConsumptionMetricsCheck                       # PAAS-2658, PAAS-2677, PAAS-2678, PAAS-2688
   - addTomcatTempCleanup                             # PAAS-2392
 
   # Actions that require a restart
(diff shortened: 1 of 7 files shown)


packages/jahia/migrations/migrate-to-v38.yaml
```diff
@@ -219,6 +219,26 @@ actions:
...
         chown dd-agent: -R $dd_checks_dir $as_conf_dir $healthcheck_conf_dir $custom_metrics_conf_dir
         systemctl restart datadog-agent
+    - cmd [storage]: |-
+        dd_dir=/etc/datadog-agent
+        dd_checks_dir=$dd_dir/checks.d
+        check_name=storage_jahia_custom_metrics
+        dd_conf_dir=$dd_dir/conf.d/${check_name}.d
+        sudoers=/etc/sudoers.d/jahia-rules
+
+        grep -q JAHIA_DD_AGENT_CHECK sudoers || \
```
julien-gagnon, line 229: I think the $ is missing for the variable sudoers

julien-gagnon (approved):

PR #940, opened 2024-02-21
PAAS-2833: change jcr path to retrieve ES password in migration packa…

Ronan-Gorain:
…ge to v38

JIRA issue: https://jira.jahia.org/browse/PAAS-2833

Short description:


kevinvandel (approved):
Approved because we need to, even though it makes no sens :grimacing:

PR #965, opened 2024-03-13
PAAS-2861: add multithreading when stopping/starting nodes during hn upgrade

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-2861

### DIFF
--- a/assets/common/hn_upgrade.py
+++ b/assets/common/hn_upgrade.py
@@ -7,6 +7,7 @@ import json
 import requests
 import datetime
 from http import HTTPStatus
+from multiprocessing import Pool
 from os import environ, path
 from pylastic import Jelastic
 from check_before_hn_upgrade import check_hardwarenodes, \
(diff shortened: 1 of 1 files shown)


assets/common/hn_upgrade.py
```diff
@@ -184,33 +196,31 @@ def get_jelastic_login_from_envname(self, envname):
...
-                e = res["response"]["error"]
-                errormsg = f'The package returned an error in {a} action: {e}'
-                logger.error(errormsg)
-                self.errors_list.append(errormsg)
-                exit(1)
+        envname = running_containers[0]
+        containers = running_containers[1]
+        logger.info("Stopping the following containers for " + envname + ": " + str(containers))
+        settings = {"nodesToStop": json.dumps(containers), "dryRun": self.dry_run}
+        res = json.loads(self.run_jelastic_package(settings, package, envname))
```
Ronan-Gorain, line 203: Shouldn't it be self.STOP_NODES_PACKAGE ?

Ronan-Gorain (changes requested):

Ronan-Gorain (approved):

PR #979, opened 2024-03-26
PAAS-2880: mute/unmute monitors impacted by other nodes in stop/start…

Ronan-Gorain:
… node packages

JIRA issue: https://jira.jahia.org/browse/PAAS-XXX

Short description:


packages/common/stop-nodes.yaml
```diff
@@ -24,6 +24,8 @@ globals:
 onInstall:
   - getNodesListsAndCheckData
   - muteNodesInDatadog
+  - muteMonitorsImpactedByOtherNodes:
+      duration: 120
```
kevinvandel, line 28: I think we need to mute env-wide monitors too

kevinvandel (changes requested):

kevinvandel (approved):

PR #981, opened 2024-03-26
PAAS-2883 patch spring-web lib

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-2883

Short description:

### DIFF
--- a/packages/one-shot/paas-2883-spring-web-patch.yaml
+++ b/packages/one-shot/paas-2883-spring-web-patch.yaml
@@ -1,0 +1,57 @@
+---
+type: update
+version: 1.5.2
+name: PAAS-2883 - Apply spring-web custom patch
+id: paas-2883-apply-spring-web-custom-patch
+description:
+  short: Update sprint-web lib to fix security issue https://academy.jahia.com/cms/render/default/customer-center/jahia/patches/security-patch-march-2024
+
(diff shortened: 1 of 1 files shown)


packages/one-shot/paas-2883-spring-web-patch.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+actions:
+  installPatch:
+    - cmd[${this}]: |-
+        patch_file="/opt/tomcat/webapps/ROOT/WEB-INF/lib/spring-web-3.2.18.jahia4_OSGI.jar"
+        if [ -f $patch_file ]; then
+          echo "[INFO] Patch is already there. Nothing to do."
+          exit 0
+        fi
+        systemctl stop tomcat
```
Ronan-Gorain, line 39: Shouldn't we use queryShutdownEndpoint action to stop it in a clean way ?

laurentfufu: indeed!

Ronan-Gorain: I'm afraid nodes will be stopped even if already fixed, and sometimes not started

Ronan-Gorain (approved):

PR #987, opened 2024-03-28
PAAS-2869: fix jcustomer 2.4 lost aliases bug at restoration

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-2869

Short description:

### DIFF
--- a/packages/common/restore.yml
+++ b/packages/common/restore.yml
@@ -808,11 +808,13 @@ actions:
 
         # Check if we restore a backup of an existing env present on the same EC deployment
         # If so, aliases will be excluded from restore to prevent conflicts errors
-        prefix=$(curl -s -u $__secret__ec_admin_credentials -XGET "https://$UNOMI_ELASTICSEARCH_ADDRESSES/_snapshot/${env.shortdomain}/${timestamp}_manual"|jq .snapshots[0].indices[0] -r | sed 's/\(.*__jc\).*/\1/')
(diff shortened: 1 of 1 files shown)


packages/common/restore.yml
```diff
@@ -808,11 +808,13 @@ actions:
 
         # Check if we restore a backup of an existing env present on the same EC deployment
         # If so, aliases will be excluded from restore to prevent conflicts errors
-        prefix=$(curl -s -u $__secret__ec_admin_credentials -XGET "https://$UNOMI_ELASTICSEARCH_ADDRESSES/_snapshot/${env.shortdomain}/${timestamp}_manual"|jq .snapshots[0].indices[0] -r | sed 's/\(.*__jc\).*/\1/')
+        prefix=$(curl -s -u $__secret__ec_admin_credentials -XGET "https://$UNOMI_ELASTICSEARCH_ADDRESSES/_snapshot/${env.shortdomain}/${timestamp}_manual"|jq ".snapshots[0].indices[0]" -r | sed 's/\(.*__jc\).*/\1/')
```
laurentfufu, line 811:
```suggestion
        prefix=$(curl -s -u $__secret__ec_admin_credentials -XGET "https://$UNOMI_ELASTICSEARCH_ADDRESSES/_snapshot/${env.shortdomain}/${timestamp}_${settings.backtype}" | jq '.snapshots[0].indices[0]' -r | sed 's/\(.*__jc\).*/\1/')
```

The important one is the backup type of course, the _quote_ things is more for readability

laurentfufu (changes requested):

packages/common/restore.yml
Ronan-Gorain: ouch... fixed

laurentfufu (approved):

PR #996, opened 2024-04-25
PAAS-2911 add one-shot package for fixing the storage nodes cloudlets

laurentfufu:
JIRA issue: https://jira.jahia.org/browse/PAAS-2911

Short description:

### DIFF
--- a/packages/one-shot/paas-2911-fix-storage-nodes-cloudlets.yaml
+++ b/packages/one-shot/paas-2911-fix-storage-nodes-cloudlets.yaml
@@ -1,0 +1,79 @@
+---
+type: update
+version: 1.5.2
+name: PAAS-2911 - Fix storage nodes cloudlets
+id: paas-2911-fix-storage-node-cloudlets
+description:
+  short: Set the correct cloudlets values for storage nodes if needed
+
+mixins:
+  - "../../mixins/common.yml"
+
+onInstall:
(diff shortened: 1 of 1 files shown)


packages/one-shot/paas-2911-fix-storage-nodes-cloudlets.yaml
```diff
@@ -0,0 +1,79 @@
...
+      nodeType: storage
+      fixedCloudlets: 0
+      flexibleCloudlets: ${globals.wantedStorageCloudlets}
+
+  decrease:
+    - log: this is the decrease operation
+    - muteDatadogHost:
+        target: storage
+        duration: 6
+    - forEach(nodes.storage):
```
Ronan-Gorain, line 65: Shouldn't jahia should be stopped/started in case of single node to avoid potential issues due to temporary FS volume loss ?

laurentfufu: Not sure about this because restart is so fast (~5s) that in my tests with a single node storage, access to the digitall site was always ok and I didn't see any error or warning in the tomcats logs. So I decided to go without,  but that's open to debate :)

Ronan-Gorain: Oh, right, no need to take care of this then

PR #1021, opened 2024-05-29
PAAS-2939: Update app-shell module

julien-gagnon:
JIRA issue: https://jira.jahia.org/browse/PAAS-2939

Short description: Update app-shell module to 2.9.0 for Jahia 8.1.3 to 8.1.6

### DIFF
--- a/packages/jahia/migrations/migrate-to-v41.yaml
+++ b/packages/jahia/migrations/migrate-to-v41.yaml
@@ -29,7 +29,7 @@ onInstall:
   ### End pre-migration actions
 
   # Actions that update modules (needs to be done first)
-  # (None)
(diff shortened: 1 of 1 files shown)


packages/jahia/migrations/migrate-to-v41.yaml
```diff
@@ -34,6 +34,7 @@ onInstall:
   # Regular actions
   - removeOldModuleVersions   # PAAS-2926
   - updateJahiaOverride       # PAAS-2941
+  - updateAppshell            # PAAS-2939
```
kevinvandel, line 37: needs to be moved to "Actions that update modules" section

kevinvandel (changes requested):

PR #1045, opened 2024-06-26
PAAS-2971: split v41 migration into 3 migration packages

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-2971

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v42.yaml
+++ b/packages/jahia/migrations/migrate-to-v42.yaml
@@ -1,0 +1,149 @@
+---
+type: update
+version: 1.5.2
+name: Migrate Jahia env to v42
+id: migrate-jahia-env-v42
+
+# Required for healthchecks
+mixins:
+  - "../../../mixins/common.yml"
+  - "../../../mixins/jahia.yml"
+  - "../../../mixins/mariadb.yml"
+  - "../../../mixins/haproxy.yml"
+  - "../../../mixins/elasticsearch.yml"
+
+globals:
+  version: 42
+
+onInstall:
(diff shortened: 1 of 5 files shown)


kevinvandel (changes requested):
As said in the ticket:
> don't forget to update the "env" version number in the Jahia install package, it should be 43 in the end

Also, we can remove the "eventsUpdate" in packages 42 and 43, but if we do so I fear we forget to add is back in the next migration package...

Ronan-Gorain:
Should be ok now

packages/jahia/migrations/migrate-to-v42.yaml
```diff
@@ -0,0 +1,149 @@
...
+            jps: "${baseUrl}../redeploy-galera-nodes.yml"
+    - else:
+        - setGlobals:
+            targetDockerTag: 10.11.7-almalinux-9
+        - getGaleraNodeNameIndex: ${nodes.sqldb.first.id}
+        - muteDatadogHost:
+            target: "*"
+            duration: 60
+        - muteEnvWideMonitors
+        - stopJahia: cp, proc
```
kevinvandel, line 138: Is it normal that we stop but never seem to restart?

kevinvandel (changes requested):

packages/jahia/migrations/migrate-to-v42.yaml
kevinvandel: I just tested it, the package fails because tomcat isn't restarted

kevinvandel (approved):

PR #1065, opened 2024-07-10
PAAS-3001: set tags before starting dd agent

julien-gagnon:
JIRA issue: https://jira.jahia.org/browse/PAAS-3001

Short description: Run tag script before starting dd agent on jcustomer

### DIFF
--- a/mixins/jcustomer.yml
+++ b/mixins/jcustomer.yml
@@ -108,7 +108,13 @@ actions:
         if [ -f $file ]; then
           mv $file $file.disabled
         fi
-        systemctl restart datadog-agent
+        # Launch "set_dd_tags" script
+        output="$(/usr/local/bin/set_dd_tags.sh)"
+        # Force agent restart in case "set_dd_tags" didn't already do it
+        if [ "$output" = "Tags are up to date, exiting." ]; then
+          echo "Force datadog-agent restart"
+          systemctl restart datadog-agent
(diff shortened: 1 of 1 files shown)


mixins/jcustomer.yml
```diff
@@ -108,6 +108,7 @@ actions:
         if [ -f $file ]; then
           mv $file $file.disabled
         fi
+        /usr/local/bin/set_dd_tags.sh
         systemctl restart datadog-agent
```
Ronan-Gorain, line 112: Maybe check the output of the script to know if restart needed or not. To avoid 2 restart in a row

julien-gagnon: Thank you for the astute comment :) Added a check

laurentfufu: I don't quite understand the point, isn't just running `set_dd_tags.sh` (and therefore removing the `systemctl restart`) enough?

julien-gagnon: My understanding is that the cron might already have run so a restart might still  be necessary to take into account the modification to the config files.

Ronan-Gorain: I understood like @julien-gagnon

laurentfufu: indded, my bad :)

Ronan-Gorain (approved):

PR #1066, opened 2024-07-11
PAAS-3004: check if SAM is started after refresh

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-3004

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -1381,6 +1381,27 @@ actions:
         # started later on anyway
         exit 0
 
+  refreshSAM:
+    # Parameter:
+    #   - target: nodegroup or nodeId
+    - log: "refresh server-availability-manager on ${this.target}"
+    - set:
+        lineNumberFile: /tmp/last_line_catalina_out
+    - cmd [${this.target}]: |-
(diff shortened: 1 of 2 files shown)


mixins/jahia.yml
```diff
@@ -1381,6 +1381,25 @@ actions:
...
+        target: ${this.target}
+        payload:
+          - karafCommand: "bundle:refresh server-availability-manager"
+    - cmd [${this.target}]: |-
+        pattern="Received event STARTED for bundle server-availability-manager"
+        log_file=/opt/tomcat/logs/catalina.out
+        line_nb=$(cat ${this.lineNumberFile})
+        timeout 60s grep -q "$pattern" <(tail -n +$line_nb -f $log_file) || \
+          { echo "timed out" && exit 0; }
+
```
kevinvandel, line 1402: Like I suggested in the ticket, I think having a log in case of timeout would be handier than only the "timed out" output to find the info from the Jelastic console logs:
```suggestion
    - if ("${response.out}"):
      - log: "[WARNING] Timeout reached when waiting for server-availability-manager start"
```

julien-gagnon (approved):

PR #1076, opened 2024-07-22
PAAS-3007: add migrate node package

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-3007

Short description:


packages/common/migrate-node.yml
```diff
@@ -0,0 +1,245 @@
...
+
+onInstall:
+  - checkCompatibilityAndDetermineNodeAndEnvType
+  - logEvent:
+      target: ${settings.nodeId}
+      title: "Restarting node ${settings.nodeId}"
+      text: "${settings.nodeId} node is going to be restarted"
+  - muteDatadogHost:
+      target: ${settings.nodeId}
+      duration: 20
```
laurentfufu, line 52: I remember (during COPS-950) that sometimes nodes migrations could last much longer than 20 minutes (even if I don't have notes on the subject), I think we shouldn't hesitate to put 60minutes here.

laurentfufu (approved):

packages/common/migrate-node.yml
Ronan-Gorain: You are right. I set 60 minutes

laurentfufu (approved):

PR #1116, opened 2024-08-30
PAAS-2584: set 60% ratio for ram/XMX

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-2584

Short description:

### DIFF
--- a/packages/jahia/migrations/migrate-to-v44.yaml
+++ b/packages/jahia/migrations/migrate-to-v44.yaml
@@ -38,6 +38,7 @@ onInstall:
   - addClassesDumps                                   # PAAS-3033
 
   # Actions that require a restart
+  - setXmxRamRatio                                    # PAAS-2584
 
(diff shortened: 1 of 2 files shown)


packages/jahia/migrations/migrate-to-v44.yaml
```diff
@@ -35,6 +35,7 @@ onInstall:
   - addCustomSshConfForKarafConsole                   # PAAS-3008
   - updateStorageNodesCloudlets                       # PAAS-2673
   - enablePageComposerForJahia82                      # PAAS-3012
+  - setXmxRamRatio                                    # PAAS-2584
```
kevinvandel, line 38: needs to be moved to "Actions that require a restart"

kevinvandel (changes requested):

kevinvandel (approved):

PR #1137, opened 2024-09-11
PAAS-1898: clean lock files during Jahia startup

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-1898

### DIFF
--- a/packages/jahia/migrations/migrate-to-v44.yaml
+++ b/packages/jahia/migrations/migrate-to-v44.yaml
@@ -378,6 +378,17 @@ actions:
         fi
         exit 1
 
+  cleanLockFilesAtJahiaStartup:
+    - cmd [cp, proc]: |-
+        filename=override.conf
+        f=/etc/systemd/system/tomcat.service.d/$filename
+        if [[ $(md5sum $f | awk '{print $1}') == "59973da5a1ab4e89026641c672531735" ]] $f; then
+          echo "Nothing to do"
+          exit 0
+        fi
+        curl -fLSso $f ${globals.repoRootUrl}/packages/jahia/migrations/v${globals.version}_assets/$filename && \
+        systemctl daemon-reload
+
 settings:
   fields:
     - type: toggle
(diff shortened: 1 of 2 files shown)


packages/jahia/migrations/migrate-to-v44.yaml
```diff
@@ -353,6 +353,17 @@ actions:
...
+  cleanLockFilesAtJahiaStartup:
+    - cmd [cp, proc]: |-
+        line='ExecStartPre=/usr/bin/bash -c '\''for dir in default live; do f="/data/digital-factory-data/repository/workspaces/$dir/locks"; [ -s $f ] || continue; until { tail -1 $f | grep -qE "^\\w*-\\w*-\\w*-\\w*-\\w*-\\w$"; }; do sed -i "$ d" $f; done; done'\'
+        f=/etc/systemd/system/tomcat.service.d/override.conf
+        if grep -q locks $f; then
+          echo "Nothing to do"
+          exit 0
+        fi
+        sed -i "/jelenv/a $line" $f
+        systemctl daemon-reload
```
kevinvandel, line 365: Wouldn't it be more reliable to check the checksum of the "override.conf" file and overwrite it completely if not up-to-date?

laurentfufu: I would also prefer :grin:

packages/jahia/migrations/v44_assets/override.conf
```diff
@@ -0,0 +1,13 @@
+[Service]
+PermissionsStartOnly=true
+EnvironmentFile=-/.jelenv
+ExecStartPre=/usr/bin/bash -c 'for dir in default live; do f="/data/digital-factory-data/repository/workspaces/$dir/locks"; [ -s $f ] || continue; until { tail -1 $f | grep -qE "^\\w*-\\w*-\\w*-\\w*-\\w*-\\w$"; }; do sed -i "$ d" $f; done; done'
```
laurentfufu, line 4: I would prefer something with a sed like this: `sed -r '${/^(\w+-){5}\w+$/!d}'`

laurentfufu (approved):

PR #1172, opened 2024-10-10
PAAS-3135: fix GraalVM conf for Jahia 8

adoux:
JIRA issue: https://jira.jahia.org/browse/PAAS-31315

### DIFF
--- a/packages/one-shot/paas-3135-fix-graalvm-and-setenv-symlink.yaml
+++ b/packages/one-shot/paas-3135-fix-graalvm-and-setenv-symlink.yaml
@@ -1,0 +1,26 @@
+---
+type: update
+version: 1.5.2
+name: PAAS-3135 - fix GraalVM and the setenv.sh symlink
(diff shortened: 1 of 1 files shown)


packages/one-shot/paas-3135-fix-graalvm-and-setenv-symlink.yaml
```diff
@@ -0,0 +1,23 @@
...
+type: update
+version: 1.5.2
+name: PAAS-3135 - fix GraalVM and the setenv.sh symlink
+id: paas-3135-fix-graalvm-and-setenv-symlink
+
+onInstall:
+  - cmd[cp, proc]: |-
+      setenv=/opt/tomcat/bin/setenv.sh
+      tomcatenv=/opt/tomcat/conf/tomcat-env.sh
+      if ! stat $setenv | grep -q "symbolic link"; then
```
Ronan-Gorain, line 11: If there is a discrepency between the wrong file and the right one, shoudln't we trigger a restart ?

PR #1175, opened 2024-10-22
PAAS-3109: stop handling robots.txt for technical domains in haproxy

Ronan-Gorain:
JIRA issue: https://jira.jahia.org/browse/PAAS-3109

Short description:


packages/jahia/migrations/migrate-to-v45.yaml
```diff
@@ -32,13 +32,14 @@ onInstall:
   # Actions that update modules (needs to be done first)
 
   # Regular actions
+  - removeRobotsTXTFromHaproxyConfig              # PAAS-3119
```
julien-gagnon, line 35: Should be PAAS-3109 instead of 3119

Ronan-Gorain: Good catch.
Fixed

laurentfufu (approved):

PR #1177, opened 2024-10-22
PAAS-3147: lower backup impact on performances and run a GC.run befor…

Ronan-Gorain:
…e data sync

JIRA issue: https://jira.jahia.org/browse/PAAS-3147

Short description:

### DIFF
--- a/packages/common/backup.yml
+++ b/packages/common/backup.yml
@@ -184,9 +184,14 @@ actions:
         DS_folder=/share/datastore
         mirror_folder=~/share_mirror
 
+        if (grep -q '^DX_VERSION=8\.1\.' /.jelenv); then
+          # Jahia 8.1
+          jcmd $(pgrep -u tomcat) GC.run
+        fi
+
         # mirroring the datastore from the share to a local directory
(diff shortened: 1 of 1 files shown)


packages/common/backup.yml
```diff
@@ -184,9 +184,14 @@ actions:
         DS_folder=/share/datastore
         mirror_folder=~/share_mirror
 
+        if [[ "$(grep DX_VESTION /.jelenv | sed 's/.*=\(8\.1\)\..*/match/')" = "8.1" ]]; then
```
julien-gagnon, line 187: Typo here, should be DX_VERSION

laurentfufu:
```suggestion
        if (grep -q '^DX_VERSION=8.1.' /.jelenv); then
```
looks better and I guess more functional than substite `8.1.*` to `match` and then test if if equal to `8.1` :raised_eyebrow:

laurentfufu (changes requested):

packages/common/backup.yml
Ronan-Gorain: indeed, don't know why first thought about the complicated way...

laurentfufu (approved):

PR #1189, opened 2024-10-30
PAAS-3168: Add one-shot to update modules

julien-gagnon:


packages/one-shot/paas-3168-update-modules-sec-104.yaml
```diff
@@ -0,0 +1,68 @@
...
+version: 1.5.2
+name: PAAS-3168 - Update modules following SEC-104
+id: paas-3168-updates-modules-sec-104
+
+mixins:
+  - "../../mixins/jahia.yml"
+  - "../../mixins/common.yml"
+
+onInstall:
+  updateModules:
```
kevinvandel, line 12: Does it really execute the "updateModules" action? I find it confusing, it looks like you define an action but you never call it. But since it's done inside the `onInstall`, I don't really know what it does.
Anyway, to make it clearer, could you please just remove the "updateModules" and put eveyrthing direcly inside the onInstall?

kevinvandel (changes requested):
We need to make sure the modules updates don't break other module, so please add modules dumps and check module state (without forgetting to exclude _all_ the module we are updating from the check :wink:)

packages/one-shot/paas-3168-update-modules-sec-104.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - if ("${globals.runningVersion}" != "${this.version}"):
+          - installOrUpgradeModules:
+              modules: jcontent/${this.version}
+      # Finish with graphql-dxm-provider if 2.X installed
+      - checkModule:
+          moduleSymname: graphql-dxm-provider
+      - if (${fn.compare([globals.runningVersion], 3.0.0)} < 0):
+          - if ("${globals.runningVersion}" != "2.21.0"):
+              - installOrUpgradeModules:
+                  modules: graphql-dxm-provider/2.21.0
```
kevinvandel, line 68: When updating graphql-dxm-provider, we must temporarily bypass SAM, like done in the update-graphql-dxm-provider package

kevinvandel (changes requested):

PR #1190, opened 2024-10-30
PAAS-3173: make sure migrations are done properly during jcustomer upgrade

adoux:
Short description:

### DIFF
--- a/packages/jcustomer/upgrade-to-v2.yaml
+++ b/packages/jcustomer/upgrade-to-v2.yaml
@@ -54,6 +54,10 @@ onInstall:
       service karaf stop
 
   - upgradeJcustomerOnNode: ${nodes.cp.first.id}
+  # If the Unomi migrations failed, the node won't start (and we shouldn't move forward with the upgrade)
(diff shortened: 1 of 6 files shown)


packages/jcustomer/upgrade-to-v2.yaml
```diff
@@ -54,6 +54,11 @@ onInstall:
       service karaf stop
 
   - upgradeJcustomerOnNode: ${nodes.cp.first.id}
+  # We need to make sure that migrations are done and the node is started
+  # before proceeding with the next one
+  - checkJcustomerHealthWhenStarting:
+      target: ${nodes.cp.first.id}
+      timeoutInMinutes: 1440
```
laurentfufu, line 61: As this is exactly the maximum execution time of a manifest on our infra, I find this value excessive as it will never be reached.
How long can it take on a send with a lot of data? If I had to guess, I'd hazard a guess that a timeout of between 5 and 8 hours would already be a more reasonable target?

laurentfufu (approved):

PR #1218, opened 2024-12-02
PAAS-3187: Install jExperience 3.5.1

julien-gagnon:
Short description:

### DIFF
--- a/mixins/jahia.yml
+++ b/mixins/jahia.yml
@@ -1905,14 +1905,17 @@ actions:
   # jexperience related #
   #######################
   getJexperienceVersion:
-    # Get jExperience version to install according to jCustomer env version:
+    # Get jExperience version to install according to Jahia / jCustomer env version:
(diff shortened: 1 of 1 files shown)


mixins/jahia.yml
```diff
@@ -1926,11 +1927,19 @@ actions:
...
             jexperienceDashboardsVersion: 1.0.0
+        - getJahiaVersion
+        - isVersionHigherOrEqual:
+            a: ${globals.jahiaVersion}
+            b: 8.1.5.0
+            res: jahia815plus
+        - isVersionHigherOrEqual:
+            a: ${this.unomiVersion}
+            b: 2.5.0
+            res: isHigherOrEqual250
```
kevinvandel, line 1938: technically it's not necessary to test this, as we don't have jCustomer versions between 2.0 and 2.4, but maybe it's more rigourous

kevinvandel (approved):

PR #1243, opened 2025-01-16
PAAS-3114: Use Datadog v2 endpoint to mute/unmute monitors

julien-gagnon:
Short description:
- Modify the way we mute monitors by tag to use the v2/downtime Datadog endpoint
- Improve env-wide muting granularity during upgrades


Ronan-Gorain (commented):
Everything done in the PR looks ok
However, the ticket talks about "mute relevant monitors depending on what nodegroup is impacted , for instance only those related to DB nodes when we redeploy a DB node."
I'm not sure it's part of the PR.

julien-gagnon:
> Everything done in the PR looks ok However, the ticket talks about "mute relevant monitors depending on what nodegroup is impacted , for instance only those related to DB nodes when we redeploy a DB node." I'm not sure it's part of the PR.

The ticket was quite confusing, as discussed. Added this part nonetheless.

Ronan-Gorain (approved):

